Issue #7, opened 2018-02-04
Include drag and drop file method


rtgdk:
Drag and drop file method to upload would make the UX more friendly. The user should also be able to use default browse-and-select file by clicking on the drag and drop tile.

You can take inspiration from gmail's drive upload button.
![capture](https://user-images.githubusercontent.com/17159160/35776016-73067cb6-09ba-11e8-9018-eaab7ee1e0cb.PNG)

freddy28:
@rtgdk I am on this one

rtgdk:
@freddy28 Sure. Go ahead. 👍

techytushar:
if anyone is not working, i would like to work on this. :blush:

rtgdk:
@freddy28  Have started some work on it but no update in the last 12 days.
Although there is much work to be done on his PR, you can check his work out.
@techytushar  If you plan to implement this feature in any other way, you are welcome to do so.

freddy28:
@rtgdk Please review #12 . It contains the first version to solving this issue.

PR #12, opened 2018-02-15
Added drag and drop capabilities to document uploading 

freddy28:
@rtgdk @pombredanne @goneall 
Added drag and drop capabilities to document uploading. 
ISSUE_FIXED: Issue [#7 ](https://github.com/spdx/spdx-online-tools/issues/7)
Please review and let me know about your ideas. 

Thanks


goneall:
@freddy28 I noticed a few binary files in the commit - are those binaries necessary?  If not, could you remove from the commit?

rtgdk:
@freddy28  You might have noticed that the tool is not functional when we upload the file by drag and drop. You need to integrate the pre-upload feature of dropzone(uploading before submitting the form) with the submit form button so that proper dialog box is opened. Also, when we upload a file through the browse feature, convert and compare feature don't work.
Also, just to make it easy and simple, if we are able to implement the drag-drop feature we wouldn't need two separate form submission on convert page. The validate feature works but we would need to show the result in the dialog box.
I would suggest, you look into how to stop the pre-upload feature and then integrate the file linked by dropzone to the file used by views function. The views function would be called by clicking on the submit button only. It should not be called as soon as the file is uploaded. So change this feature too.
Also, remove unnecessary files before commiting.

pombredanne:
@freddy28  Can you add a DCO signoff to your commit?
See https://developercertificate.org/
And also remove the duplication in your commit message?

rtgdk:
@freddy28  Please remove unnecessary binary files from commit, also update your fork master with the current master branch to resolve the merge conflicts.

rtgdk:
Work completed in #28 . Closing this PR.

PR #31, opened 2018-06-12
License submittal

rtgdk:


src/app/templates/app/submit_new_license.html
```diff
@@ -0,0 +1,112 @@
...
+          <label class="control-label" id="shortidentifierlabel" ></label>
+        </div>
+      </div>  
+    </div>
+
+    <div class = "form-group">
+      <div class="col-sm-12">
+      <label class="control-label col-sm-3" >Source / URL</label>
+        <div class="col-sm-3" >
+          <button class="roundedButton btn btn-md btn-info" id="addSourceUrl" type="submit" >+</button>
```
rtgdk: This button does not work. The page refreshes without displaying any changes.

src/app/templates/app/submit_new_license.html
```diff
@@ -0,0 +1,112 @@
...
+        <div class="col-sm-3">           
+          <textarea form ="newlicense" name="text" id="text" cols="40" wrap="soft" rows="3"></textarea>
+          <label class="control-label" id="noteslabel" ></label>
+        </div>
+      </div>  
+    </div>
+
+    <hr>
+		<input type="hidden" id="cfileformat" name="submit" value="">
+		<button class=" btn btn-md btn-info" id="submitbutton" type="submit" >Submit</button>
```
rtgdk: Submit button doesn't save any data. Have you not implemented the database yet?

rtgdk: Make the submit button work through AJAX. You can look at other pages like validate.html  how they send requests. Also, display warnings/error if any. 
This AJAX will also be useful to send request to GitHub API.

src/app/views.py
```diff
@@ -53,6 +53,24 @@ def about(request):
         'app/about.html',context_dict
         )
 
+def submitNewLicense(request):
+    """ View for submit new licenses
+    returns submit_new_license.html template
+    """
+    context_dict={}
+    return render(request,
```
rtgdk: Please write the code to save the submitted request to the database.

rtgdk:
@gaalocastillo  Please look at the above suggested changes. You should speed up your work. You have missed many important initial things like saving the form submitted, proper functioning button.

rtgdk:
@gaalocastillo  You have not committed any changes till now. You should spend at least 30-35 hours per week on your GSoC project according to the official guidelines. I am not sure whether you are giving that much amount of time to this project.  
From now on, I want to see **daily commits from you on this project.** Also, please **post your project update on the gitter channel daily.** 

If you are facing any issue regarding time management or coding in python/javascript, please let us know.

src/app/templates/app/submit_new_license.html
gaalocastillo: You can check the AJAX request implementation including the warnings I just commited @rtgdk .

src/app/models.py
```diff
@@ -11,18 +11,30 @@
...
-    user = models.OneToOneField(User)
-    organisation = models.CharField("Organisation",max_length=64, null=False, blank=False)
-    lastlogin = models.DateField("Last Login",default=datetime.now,blank=True)
-    def __str__(self):
-        return self.user.username
+	user = models.OneToOneField(User)
+	organisation = models.CharField("Organisation",max_length=64, null=False, blank=False)
+	lastlogin = models.DateField("Last Login",default=datetime.now,blank=True)
+	def __str__(self):
+		return self.user.username
```
rtgdk: Please use the default 4 space indentation instead of tab. No need to change already written code.

src/app/models.py
```diff
@@ -11,18 +11,30 @@
...
+
+class LicenseRequest(models.Model):
+	fullname = models.CharField(max_length=70)
+	shortIdentifier = models.CharField(max_length=25)
+	submissionDatetime = models.DateTimeField(auto_now_add=True)
+	userEmail = models.EmailField(max_length=35)
+
```
rtgdk: Why aren't you adding all the fields in the models like you did in forms? Also, if you added all the field in the models, you can simply use `class Meta:` in forms. This will be a much more efficient way. Also, if there are any errors to the form submitted, you can easily check for them in forms.py . Let me know if you need help in this.

src/app/views.py
```diff
@@ -53,6 +55,44 @@ def about(request):
...
+    """ View for submit new licenses
+    returns submit_new_license.html template
+    """
+    if request.method == 'POST':
+        form = LicenseRequestForm(request.POST, auto_id='%s')
+        if form.is_valid():
+            # process the data in form.cleaned_data as required
+            # ...
+            # redirect to a new URL:
+#            return HttpResponseRedirect('/submit_new_license/')
```
rtgdk: Getting an error here for indentation. Please fix.

rtgdk: Also, as @goneall  suggested don't comment out old code. This made the `if form.is_valid():` block without any statement.

rtgdk:
@gaalocastillo  I am not sure why you changed database settings from sqlite to postgres. For working locally I think sqlite is the way to go. Postgres can be set up later in production. Please switch back to sqlite. If you have other reasons for using postgres, please let me know.

Also, I have added some review comments, please look at them and fix.

src/src/settings.py
```diff
@@ -81,8 +81,12 @@
 
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
```
rtgdk: I think sqlite3 will work for your project.

src/app/views.py
```diff
@@ -53,6 +55,44 @@ def about(request):
...
+    else:
+        form = LicenseRequestForm(auto_id='%s')
+    context_dict={'form': form}
+    return render(request, 
+        'app/submit_new_license.html', context_dict
+        )
+
+def generateLicenseXml(licenseOsi, licenseIdentifier, licenseName, licenseSourceUrls, licenseHeader, licenseNotes, licenseText):
```
rtgdk: This function is never called. Also I think you can use the `if form.is_valid():` block in submitNewLicense function for saving the model

rtgdk: Let me know if this function does something different or if you plan to do something different with this function later on.

goneall:
I can add the PostgreSQL database in when we deploy.

We could have 2 config files - one for production and one for development so we can track the changes.

src/app/views.py
gaalocastillo: This function is the one that will be called to generate the xml file. I was going to call it inside the submitNewLicense function, but if you suggest I can implement it directly in submitNewLicense.

gaalocastillo:
I am going to fix the changes suggested @rtgdk . I have no problem working with sqlite, so I can fix it. I was thinking in start using it from now but it is ok to keep working with sqlite.

src/app/views.py
rtgdk: I thought this would save the model. But if this is generating xml files, then it would be best to keep this as a separate function. But then why aren't you saving the form in submitNewLicense function?

rtgdk:
@gaalocastillo  Any updates on the changes? 
Regular updates/commits would be best.

gaalocastillo:
The license requests submitted are stored on database.

- The form fields still need to be empty after submission.
- Only one license url is included in the license xml file.
- GitHub issue submission is not included yet.

@rtgdk please let me know if I am missing anything else about this feature, considering what is already implemented in the _submitNewLicense_ view by now.

src/src/settings.py
gaalocastillo: Sqlite is back again on the database, I agree with you.

gaalocastillo:
I have created an file to automatically insert LicenseRequest records into the database. This is just for  testing the functionalities easier. Do you consider good to add to the readme the command for loading the data from this file? @rtgdk @goneall 
Also, now the license requests submitted are loaded from the database, in the license requests list view.

src/src/settings.py
```diff
@@ -45,7 +45,7 @@
     'rest_framework',
 ]
 
-MIDDLEWARE = [
+MIDDLEWARE_CLASSES = [
```
rtgdk: I think for Django-1.11 it's MIDDLEWARE. Although, MIDDLEWARE_CLASSES might have backward compatibility. But why change this? I think MIDDLEWARE would also work just fine.
Also, please put a comment when you are changing something in src folder(like settings.py, urls.py, wsgi.py), since it will affect all the applications.

src/licenseRequests.json
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+]
```
rtgdk: Can you please put this file in the static folder? or Create another folder for testing purposes and add a README.md in that folder. If you require more files to be loaded for testing you can put all of them in that folder.if would be easier to distinguish and remove them after testing. Putting testing files in the main folder is not good.

src/app/models.py
```diff
@@ -11,18 +11,31 @@
...
+class LicenseRequest(models.Model):
+    fullname = models.CharField(max_length=70)
+    shortIdentifier = models.CharField(max_length=25)
+    submissionDatetime = models.DateTimeField(auto_now_add=True)
+    userEmail = models.EmailField(max_length=35)
+    xml = models.TextField()
+
+    def __unicode__(self):
+        return "%s %s %s %s %s" % (self.fullname, self.shortIdentifier, self.submissionDatetime, self.userEmail, self.xml)
```
rtgdk: You can only return the unique identifier for this.I(Like fullname or shortIdentifier). It shows like this in the admin window.
screenshot-2018-7-1 select licenserequest to change django site admin

Doesn't look good.

src/app/templates/app/license_requests.html
```diff
@@ -0,0 +1,51 @@
...
+<div class="panel panel-default">
+<div class="panel-heading"> <p class="lead">License Requests List</p> </div>
+<div class="panel-body">
+    <table id="requests_table" class="display dataTable">
+      <thead>
+        <tr>
+          <th>Fullname</th>
+          <th>Short Identifier</th>
+          <th>Submission Date</th>
+        </tr>
```
rtgdk: This page shows just these 3 variables of the submitted license request. We need to create template/url to view more info about them. Add a link to that page here.

src/app/templates/app/license_requests.html
```diff
@@ -0,0 +1,51 @@
...
+{% block script_block %}
+<script type="text/javascript">
+  $(document).ready(function () {
+      $("#licenserequestspage").addClass('linkactive');
+      });
+</script>
+
+<script type="text/javascript">
+  $(document).ready( function () {
+      $('#requests_table').DataTable();
```
rtgdk: Currently the search happens over all the fields- Like Fullname, short identifier and Date. Maybe we can add more specific search later. I think this JS datatables provides that capability. 
You can add this to the ToDo list. No need to worry about it now.

src/app/forms.py
```diff
@@ -55,4 +60,14 @@ class OrgInfoForm(forms.ModelForm):
     
     class Meta:
         model = UserID
-        fields = ('organisation',)
\ No newline at end of file
+        fields = ('organisation',)
+
+class LicenseRequestForm(forms.Form):
```
rtgdk: You can use the `class Meta:` with model `LicenseRequest`. See other forms for examples. You would not need to add forms.Charfield for every variable. Also `class Meta:` provides some inbuilt checking - like if fields are empty/unique/etc. So it would be best to use the Meta class for forms.

rtgdk:
1. Bug: When I submit a license, it displays the following modal. Once the form is successfully submitted, all the form fields should be clear.
screenshot-2018-7-1 spdx online tool 3
2. I am not sure if you should include only one url or more? @goneall can help you you with this.

rtgdk:
@gaalocastillo Also, I answered your questions in the review above. Please check.
Also, let me know if you face any difficulty with the above changes.

src/app/templates/app/license_requests.html
gaalocastillo: Yes, actually that was a resolution we got in the call in which we discussed about the design of the database. The three variables showed in the list view are just for filtering, but when user clicks on any row, it will be redirected to a new template/url wich will contain the information of the license/request.
Would you suggest to add a click listener on the entire row? @rtgdk . Or one more column having for each row a clickable "View license" message?

src/src/settings.py
gaalocastillo: I was having some problems loading the Django admin page. This error was being displayed: _'WSGIRequest' object has no attribute 'user' Django admin_; a little similar to this:  https://stackoverflow.com/questions/37949198/wsgirequest-object-has-no-attribute-user-django-admin
There I found that changing MIDDLEWARE to MIDDLEWARE_CLASSES would solve the problem without affecting in a bad way the application.

gaalocastillo: Is it ok just commenting here the changes in src folder? @rtgdk , or do you mean somewhere else? I will have in mind next time.

src/app/templates/app/license_requests.html
rtgdk: You can have a link to the template/url for the whole row. Like in HTML
`<a href="/.../{particular_license_page}"><tr>.....</tr></a>`

src/src/settings.py
rtgdk: Can you please check the exact version of Django. It should be 1.11.2. Let me know.

rtgdk: MIDDLEWARE was introduced in 1.10 version, so it would be better to keep MIDDLEWARE_CLASSES to your own local version. Can you please not commit this change, since it will affect all the other application?

src/licenseRequests.json
gaalocastillo: Great, I am going to create that folder. Thanks for the suggestion!

src/src/settings.py
gaalocastillo: It is ok, I will use it just in my local version.

src/licenseRequests.json
gaalocastillo: Fine, I am going to create the folder and add both the readme file and the testing file.

src/src/settings.py
gaalocastillo: I will add back agaim MIDDLEWARE.

src/app/forms.py
gaalocastillo: Ok, I will  look how this works. I actually do not know how this could be used, then I will see other examples.

src/app/models.py
gaalocastillo: Done. Now it just shows the license fullname.

src/api/migrations/0002_auto_20180701_0923.py
```diff
@@ -0,0 +1,84 @@
...
+            model_name='convertfileupload',
+            name='file',
+            field=models.FileField(upload_to=api.models.user_directory_path),
+        ),
+        migrations.AlterField(
+            model_name='validatefileupload',
+            name='file',
+            field=models.FileField(upload_to=api.models.user_directory_path),
+        ),
+    ]
```
rtgdk: Can you please not commit all the local migrations files. Just start the start the migrations from the beginning and just commit the 0001_initial.py.

src/app/forms.py
rtgdk: Just see the above examples of UserRegisterForm, UserProfileForm. It's right above the class you made.

src/app/migrations/0002_licenserequest.py
```diff
@@ -0,0 +1,30 @@
...
+                ('submissionDatetime', models.DateTimeField(auto_now_add=True)),
+                ('userEmail', models.EmailField(max_length=35)),
+                ('xml', models.TextField()),
+            ],
+            options={
+                'verbose_name': 'LicenseRequest',
+                'verbose_name_plural': 'LicenseRequests',
+            },
+        ),
+    ]
```
rtgdk: Again, this can be combined in the 001_initial.py migrations. Just delete your migration files and run `python manage.py makemigrations` and commit that file.

rtgdk: After that you would have to run migrate command also. ( and createsuperuser too, if you want admin access)

src/app/templates/app/submit_new_license.html
```diff
@@ -0,0 +1,221 @@
...
+    return ("Please enter the license fullname.");
+  }
+  else if (($('#shortIdentifier').val().trim() == "")){
+    $('#shortidentifier').focus();
+    return("Please enter the identifier.");
+  }
+  else if (($('#osiApproved').val() == "0")){
+    $('#osi_approved').focus();
+    return("Please select if the OSI is approved or not.");
+  }
```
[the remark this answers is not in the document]
gaalocastillo: Ok, I will do these changes in my next commits.

rtgdk:
@goneall @gaalocastillo  I am merging this PR to the license-submittal branch. For other commits, you can open a new PR if needed.

PR #38, opened 2018-07-25
GitHub issue creation

gaalocastillo:
Any suggestion please let me know @goneall @rtgdk

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -42,6 +42,8 @@ from app.forms import UserRegisterForm,UserProfileForm,InfoForm,OrgInfoForm
 from .forms import LicenseRequestForm
 from .models import LicenseRequest
 
+from utils.github_utils import getGithubToken
+
 def index(request):
     """ View for index
     returns index.html template
@@ -109,9 +111,8 @@ def createIssue(licenseName, licenseIdentifier, licenseSourceUrls, licenseOsi):
     """ View for creating an GitbHub issue
     when submitting a new license request
     """
-    myToken = ''
+    myToken = getGithubToken()
     body = '**1.** License Name: ' + licenseName + '\n**2.** Short identifier: ' + licenseIdentifier + '\n**3.** URL: '
-    print 'longitud: ' + str(licenseSourceUrls)
     for url in licenseSourceUrls:
         body += url
         body += '\n'
@@ -121,7 +122,6 @@ def createIssue(licenseName, licenseIdentifier, licenseSourceUrls, licenseOsi):
     headers = {'Authorization': 'token ' + myToken}
     url = 'https://api.github.com/repos/spdx/license-list-XML/issues'
     r = post(url, data=dumps(payload), headers=headers)
-    print r.status_code
 
 def licenseRequests(request):
     """ View for license requests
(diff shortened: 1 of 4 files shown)


src/app/views.py
```diff
@@ -103,6 +105,24 @@ def generateLicenseXml(licenseOsi, licenseIdentifier, licenseName, licenseSource
...
+        body += url
+        body += '\n'
+    body += '**4.** OSI Approval: ' + licenseOsi
+    title = 'New license request: ' + licenseIdentifier + ' [SPDX-Online-Tools]'
+    payload = {'title' : title, 'body': body, 'labels': ['new license/exception request']}
+    headers = {'Authorization': 'token ' + myToken}
+    url = 'https://api.github.com/repos/spdx/license-list-XML/issues'
+    r = post(url, data=dumps(payload), headers=headers)
+    print r.status_code
+
```
goneall: We should add some error handling here - if the status is not a success, open a dialog "Please note that there was a problem opening the issue for the SPDX legal team.  Please email spdx-legal@lists.spdx.org with SPDX ID for the license you are submitting"

src/app/views.py
```diff
@@ -103,6 +105,24 @@ def generateLicenseXml(licenseOsi, licenseIdentifier, licenseName, licenseSource
     xmlString = ET.tostring(root, encoding='utf8', method='xml')
     return xmlString
 
+def createIssue(licenseName, licenseIdentifier, licenseSourceUrls, licenseOsi):
+    """ View for creating an GitbHub issue
+    when submitting a new license request
+    """
+    myToken = ''
```
goneall: We should move the token to an external file which is not under source control for security.  There are a few methods we can use to do this.  My preference would be to create a separate Python file with myToken defined.  We would import from this file here.  Then we would add the file to the .gitignore so that the tokens do not get accidentally committed.

src/app/views.py
gaalocastillo: Ok sure, I am going to be working on this.

src/app/views.py
gaalocastillo: That solution sounds good. I will work on it.

src/app/views.py
gaalocastillo: Now a warning message is displayed when the issue cannot be created. However, even when an issue could not be created, it could be stored on the submitted requests database. What do we do with requests whose issue cannot be created? Do we still save them or not? @goneall @rtgdk

src/app/views.py
```diff
@@ -122,6 +126,10 @@ def createIssue(licenseName, licenseIdentifier, licenseSourceUrls, licenseOsi):
     headers = {'Authorization': 'token ' + myToken}
     url = 'https://api.github.com/repos/spdx/license-list-XML/issues'
     r = post(url, data=dumps(payload), headers=headers)
+<<<<<<< HEAD
+    return r.status_code
+=======
+>>>>>>> c40bd705c4a9cf86510cf680da0912e9ec650223
```
goneall: It looks like you committed some merge conflict text

src/app/views.py
goneall: I would still save the info in the DB - the legal team could follow-up and create an issue manually using the information from the DB.  We could enhance this in the future to email the legal team that an error occurred, but I think the warning message is fine for now.

goneall:
I'll go ahead and merge into the the license-submittal branch

PR #46, opened 2018-09-01
License text field format

gaalocastillo:
I have made some changes on the way the license **text field** is displayed on the license information view:
- HTML tags are supported on the license text field without creating any conflicts with the license XML creation.
- Linefeeds have been replaced with `<br>` tags.
- The unnecessary `<text>` tag surrouding the text field has been removed.

Please let me know any change/suggestion @rtgdk @goneall . I think my solution could follow cleanner approach.

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -45,6 +45,8 @@ from .models import LicenseRequest
 
 from utils.github_utils import getGithubToken
 
+import cgi
+
 def index(request):
     """ View for index
     returns index.html template
@@ -78,11 +80,6 @@ def submitNewLicense(request):
             licenseHeader = form.cleaned_data['licenseHeader']
             licenseNotes = form.cleaned_data['notes']
             licenseText = form.cleaned_data['text']
-            if '\n' in licenseText:
-                licenseText = '<![CDATA[' + licenseText
-                licenseText = licenseText + ']]>'
-                licenseText = licenseText.replace('\r', '')
-                licenseText = licenseText.replace('\n', '<br>')
             userEmail = form.cleaned_data['userEmail']
             xml = generateLicenseXml(licenseOsi, licenseIdentifier, licenseName,
                 licenseSourceUrls, licenseHeader, licenseNotes, licenseText)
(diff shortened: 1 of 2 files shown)


src/app/views.py
```diff
@@ -78,6 +78,11 @@ def submitNewLicense(request):
             licenseHeader = form.cleaned_data['licenseHeader']
             licenseNotes = form.cleaned_data['notes']
             licenseText = form.cleaned_data['text']
+            if '\n' in licenseText:
+                licenseText = '<![CDATA[' + licenseText
+                licenseText = licenseText + ']]>'
+                licenseText = licenseText.replace('\r', '')
+                licenseText = licenseText.replace('\n', '<br>')
```
goneall: If you replace the `<br>` with a `<br />` you will have valid XML

src/app/views.py
```diff
@@ -218,8 +223,14 @@ def parseXmlString(xmlString):
...
-            textStr = ET.tostring(textElem)
-            data['text'] = textStr
+            textStr = ET.tostring(textElem).strip()
+            if(len(textStr) >= 49 and textStr[:42] == '<text xmlns="http://www.spdx.org/license">' and textStr[-7:] == '</text>'):
+                textStr = textStr[42:]
+                textStr = textStr[:-7].strip().replace('&lt;', '<').replace('&gt;', '>').strip()
+            if(len(textStr) >= 12 and textStr[0:9] == '<![CDATA[' and textStr[-3:] == ']]>'):
+                textStr = textStr[9:]
+                textStr = textStr[:-3]
+            data['text'] = textStr.strip()
```
goneall: I'm not understanding the reason for adding the CDATA and removing it here.

Adding the `<br />` tags should be valid HTML.

I also tested this out and the CDATA is present in the downloaded XML.

goneall: I would treat the text entered by the user just as plain text.  We should escape the text for HTML (e.g. `<` -> `&lt;`) or any other characters which may not be parseable by the XML.

PR #48, opened 2018-09-25
Change method of formatting the text to use the paragraph elements

goneall:

Changed the method of formatting from CData to using paragraph tags in the XML editor.

Also added linefeeds around the XML tags to format a bit better - it isn't a very good pretty printer, but it looks better than no linefeeds at all.  We should change to a better XML pretty printer once everything is merged.

@gaalocastillo If you could review the changes
@rtgdk If you could also review and if it looks OK, merge this into text-field-format, then merge text-field-format into the license-submittal branch before merging with master


rtgdk:
Looks good. Merging into text-field-format

Issue #54, opened 2018-10-04
Submit new license - use github username of logged in user


goneall:
Currently, it uses the github user configured in the application (goneall)

goneall:
Based on the legal call on 1 Nov 2018, we will require a github user for the submitter.  We can use the same mechanism we use for creating the pull request to require the user to login and submit the issue from that github user's account.

techytushar:
@goneall I guess this can now be closed !!

PR #68, opened 2018-11-16
GitHub license submit

goneall:
Resolves issue #54 by authenticating the user through github before submitting a new license request.

Much of the logic and some of the code was copied from the github authentication used in the XML pull request code.

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -82,7 +82,12 @@ def submitNewLicense(request):
     ajaxdict = {}
     if request.method=="POST":
         if not request.user.is_authenticated():
-            return HttpResponseRedirect(settings.LOGIN_URL)        
+            if (request.is_ajax()):
+                ajaxdict["type"] = "auth_error"
+                ajaxdict["data"] = "Please login using GitHub to use this feature."
+                response = dumps(ajaxdict)
+                return HttpResponse(response,status=401)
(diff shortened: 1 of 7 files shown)


src/app/views.py
```diff
@@ -79,35 +78,70 @@ def submitNewLicense(request):
...
-            licenseRequest.save()
-            statusCode = createIssue(licenseName, licenseIdentifier, licenseSourceUrls, licenseOsi)
-            data = {'statusCode' : str(statusCode)}
-            return JsonResponse(data)
-
+    context_dict = {}
+    ajaxdict = {}
+    if request.method=="POST":
+        if not request.user.is_authenticated():
+            return HttpResponseRedirect(settings.LOGIN_URL)
```
rtgdk: This return is making the ajax behave unexpectedly when the user is not logged in. The correct code would be same as the expect block of UserSocialAuth.DoesNotExist:

```suggestion
            if (request.is_ajax()):
                ajaxdict["type"] = "auth_error"
                ajaxdict["data"] = "Please login using GitHub to use this feature."
                response = dumps(ajaxdict)
                return HttpResponse(response,status=401)
            return HttpResponse("Please login using GitHub to use this feature.",status=401)
```

rtgdk:
Apart from that change, everything else works like a charm.

Issue #64, opened 2018-11-01
Submit new license: change Notes field


jlovejoy:
Can we rename the "Notes" field to "Comments" and make the text box bigger, and move it to the bottom of the form?

"Notes" are an official field in the SPDX License List, but I don't think that is really needed at this point in the process.
 We do need and want to enable a place for longer comments, as people often include in the current email submission, e.g., where the license is used should be included here

techytushar:
Should the notes field be also changed on the `license_requests` page and in the downloaded XML file?

jlovejoy:
@techytushar  - good question... I'm not sure what you mean by license_requests page and in the downloaded XML file though.
The Comments in the submission form will only be needed for the review process. In other words, "comments" is not a field that maps to the final license information that is added to the SPDX License List

jlovejoy:
another alternative here would be to simply add a new field for Comments; and leave Notes and I can add documentation as to what to put there. Maybe this is easier, especially if Notes is already mapped to other output?

goneall:
I would have the "comments" just be included in the issue and the database and not include it in the XML.  I'm thinking that the comments would just be related to the submission and not necessary intended to be part of the permanent license list data for the license.  We can always manually add the comments to the existing notes field using the XML editor.

jlovejoy:
@goneall - I think I confused things here, but we are headed back around in the right direction. The point of having a "comments" (free text field) in the Submission form is to allow for any additional information about the license or why someone is requesting it, like where it is used. The current license submission form does not have room for that, but it is an important aspect. I don't think it's efficient to expect a submitter to fill out the form, then go to the issue and add comments there - that makes for two steps for the initial submission. 

you are correct that any comments to this end would NOT be included as part of the XML file that gets created if the license is accepted or part of the license list data for the license. 

does that help clarify?

goneall:
This has been merged and deployed.  Thanks @ekongobie for the contribution!

goneall:
The notes are actually being added in the XML and not in the issue - I'll open a separate issue to change this behavior.

PR #75, opened 2019-01-10
Rename Note to comment, change to textarea, move to bottom

ekongobie:
#64 
rename the "Notes" field to "Comments", make the text box bigger, and move it to the bottom of the form


rtgdk:
Looks good.

Issue #76, opened 2019-01-10
Add edit xml button on license requests page 


techytushar:
We can add a `Edit XML` button besides the `Download XML` button on the `license_requests` page which would open the XML file in the editor so that the user can edit it further. Will that be a good idea @goneall @rtgdk

goneall:
@rtgdk I need to go back through the workflow.  There are 2 scenarios, one where a member of the legal team is submitting a license request and another submitter when someone outside the legal team is submitting.  For the former, I think having a button to go directly to the XML editor would make the workflow easier and faster.  For the latter, they probably would not edit the XML since they would not be familiar with the format.

@jlovejoy As someone who has submitted a few licenses, what do you think?

jlovejoy:
so are you thinking to add a "edit XML" button on this page: http://13.57.134.254/app/license_requests/ (in the list) or on the specific page for the license itself? i.e., http://13.57.134.254/app/license_requests/11/

Would this "edit xml" button take you automatically to the XML Editor tool and functionality?

tjasmith:
@techytushar @goneall @jlovejoy 
Is it Ok if I work on this issue/enhancement? Or do you have a more pressing and important issue which you want me to work on?

techytushar:
@jlovejoy Yes, the idea is to display the `Edit XML` button on this page: http://13.57.134.254/app/license_requests/11/ which will take the user to the XML Editor with the text of that license prefilled.

goneall:
@tjasmith I think this would be a good enhancement to work on

tjasmith:
@goneall. On it !

goneall:
@tjasmith contributed a solution with PR #79 which has been deployed

PR #79, opened 2019-01-20
Add edit xml functionality on license request page 

tjasmith:
For issue #76

### DIFF
--- a/src/app/urls.py
+++ b/src/app/urls.py
@@ -38,6 +38,7 @@ urlpatterns = [
     url(r'^checkusername/$', views.checkusername, name='check-username'),
     url(r'^xml_upload/$',views.xml_upload, name='xml-upload'),
     url(r'^edit/(?P<page_id>[0-9a-z]+)/$', views.xml_edit, name='editor'),
+    url(r'^edit_license_xml/(?P<license_id>[0-9]+)/$', views.edit_license_xml, name='license_xml_editor'),
     url(r'^validate_xml/$', views.validate_xml, name='validate-xml'),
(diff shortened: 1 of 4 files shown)


tjasmith:
@techytushar @rtgdk @goneall Please do not review yet, as I am working on a way to make this solution better.

tjasmith:
@techytushar You can now review. I have updated the PR.

Thanks.

tjasmith:
@rtgdk @techytushar 
In case of an error, I redirect to the 404 page.
I have updated my branch with the latest version of the repo.
I have added tests on the edit xml functionality, and here is a screenshot of the tests as they pass:
tests2

src/app/urls.py
```diff
@@ -38,6 +38,7 @@
     url(r'^checkusername/$', views.checkusername, name='check-username'),
     url(r'^xml_upload/$',views.xml_upload, name='xml-upload'),
     url(r'^edit/(?P<page_id>[0-9a-z]+)/$', views.xml_edit, name='editor'),
+    url(r'^edit_license_xml/(?P<license_id>[0-9]+)/$', views.edit_license_xml, name='license_xml_editor'),
```
techytushar: I would suggest you to put `*` instead of `+` at that Regex, then only the redirect would work

techytushar:
Just a small change otherwise LGTM, great work :+1:

src/app/urls.py
tjasmith: @techytushar 
I have added the changes you kindly requested.

goneall:
Tested this out locally - merging into the master and deploying.

@tjasmith Thanks for the contribution!

PR #82, opened 2019-01-20
Update django version

### LINKED ISSUE #80: Update django version
There is a minor security vulnerability in django version 1.11.18.  We should upgrade to django ~> 1.11.18.

Note: related to issue #58

Ugtan:
Fixes #80

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django==1.11.16
+Django>=1.11.18,<1.12
 jpype1==0.6.3
 numpy==1.15.2
 djangorestframework==3.6.4


rtgdk:
You might to look into the changes between 1.11.16 and 1.11.18 and any existing vulnerability in the repo. (Like any function that caused the vulnerability was used in the repo or not). Also, you can look into the changes between the 2 version's settings.py file and other files in the `src` folder

Ugtan:
The 1.11.18 patch fixed the content spoofing in `django.views.defaults.page_not_found()` which is not used in the repo. I also tried running it locally with the 1.11.18 django version and it works fine. @rtgdk

rtgdk:
LGTM.

PR #88, opened 2019-01-31
Implement python decouple

### LINKED ISSUE #85: Add Python Decouple
### Description
Python Decouple strictly separates the settings parameters from the source code. It adds parameters related to an instance of the project into a seperate .env file. All the environment-specific parameters (secret key, user, passwords, debug status and hosts to mention a few) can be kept into and changed from a single configuration module. Also, It enables changing parameters without having to redeploy the app.

Adding this into .gitignore ensures these are not pushed with the code.

The following dependency is required for this change:
`python-decouple`

abhishekspeer:
Implemented python-decouple. 

All the environment-specific parameters (secret key, user, passwords, debug status etc.) are now kept into a single .env file. Also, we won't be needing  `secret.py` script anymore. Here are the changes made to achieve the same:

- Added .env script.
- Added .env to .gitignore to ensure these are not pushed with the code.
- Updated requirements.txt to add `python-decouple`.
- Made the required changes in README.md.

Fixes #85


rtgdk:
Fixed with #180. Closing this

PR #90, opened 2019-03-04
Fix button in error templates

Ugtan:
The home page button in error templates was not working, now it takes the user to the
home page after clicking on it.

### DIFF
--- a/src/app/templates/400.html
+++ b/src/app/templates/400.html
@@ -8,5 +8,5 @@
 Error 400
 </h1></p>
 <br>
-<button class="btn btn-lg btn-info"><a href="/app/" style="color:black;">Home Page </a></button>
+<a class="btn btn-lg btn-info" href="/app/" style="color:black;">Home Page </a>
 {% endblock %}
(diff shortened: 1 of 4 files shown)


rtgdk:
@Ugtan Are you sure about this? It should work fine since we are extended base html and loading all static files. If not, can you please specify which url was the button taking you to previously?

Ugtan:
You are right it should work but it was not working in mine as i am using firefox browser. After a quick search i found out that it doesn't work in firefox, you can refer [here](https://stackoverflow.com/questions/16280684/nesting-a-inside-button-doesnt-work-in-firefox). So i tried to fix it using js onclick. Another possible solution would be `<a href="#"><button class="btn"></button></a>` as mentioned in the above link.

rtgdk:
@Ugtan Okay, I was testing on Chrome. Using `<a> tag` with `class="btn btn-lg btn-info"` would be good.
Update it and I'll merge the PR.

Ugtan:
@rtgdk okay will do it soon.

Ugtan:
@rtgdk Done. Please have a look.

PR #96, opened 2019-03-16
Limit requests per hour

### LINKED ISSUE #4: Create a mechanism to limit the requests to the online tools
Programmatic access could put a substantial load on the service if multiple requests are made with high frequency.

Suggest an enhancement to govern the rate at which requests from the same user and/or IP address is made.

This could also apply to the web access.

tjasmith:
Fixes: https://github.com/spdx/spdx-online-tools/issues/4

The license sumission is reduced to 100 per hour by a particular user.
When the limit is reached, the modal below is shown.
quota2


Ugtan:
@tjasmith Why add a whole new module when you can simply do it with drf throttling?

tjasmith:
@Ugtan 
Thanks for the question.

I decided to go on with [django rate limit](https://django-ratelimit.readthedocs.io/en/stable/index.html) because it can be used on normal django views(which is the view on which it was supposed to be implemented.)
Moreover, it can be used on class-based views(if we decide to switch to those), and can also be applied with a mixin.

I read about django rest framework throttling, but it didn't suit my needs(adding this to a normal django view), while the module which I choose did.

goneall:
Will this limit the rate on all requests or just the license submittal?   The request I'm most concerned about is the [check license](http://13.57.134.254/app/check_license/).  It would be good to limit all of the requests that can be made through the API.

tjasmith:
@goneall 
This is just for the license submittal request. Ok, I shall now implement this on all other requests.

rtgdk:
@tjasmith Also, write tests to check whether this modal is actually shown or not. You can use selenium to automate these requests.

tjasmith:
@rtgdk

Ok. Thanks for the hint.

tjasmith:
@goneall I have limited the rate on all django rest framework apis, and on the following post requests:
- check license
- validate
- compare
- convert
- xml upload
- validate xml
- submit license

@rtgdk I have written tests for the return of the requests using django's RequestFactory(to be able to simulate that the rate limit has been reached).

goneall (commented):
Changes look good to me.  There are some changes in the white space which is causing some extra diffs.  Are these necessary?  I'm concerned it may create more merge conflicts.

rtgdk:
@tjasmith In the test, you should send 100 request from selenium(either one by one or a bunch of them in parallel [In different tabs/windows of the browser]). The test you have written tests whether the view function is returning the correct error or not. You can keep this test. But add the selenium test  to test whether this error is shown in real or not on the UI.

tjasmith:
@rtgdk Ok, On it.

bact:
There are a bunch of merge conflicts to be resolved before this can be merged.

rtgdk:
I don't think we need this now. Also, the core logic looks incorrect now. We can start a new PR for this. Closing this without merging.

PR #101, opened 2019-03-18
Change the comments field to report the comments in the issues

### LINKED ISSUE #98: Change the comments field to report the comments in the issues and not include in notes
Currently, when you submit a license, the comments are included in the notes section of the XML.

The legal team would like to include the comments in the Issues submitted to the license-list-XML repo and not include it in the `<notes> ... </notes>` section of the XML.  We'll reserve the notes section of the XML for legal team use.

Ugtan:
Fixes #98

### DIFF
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -79,6 +79,6 @@ class LicenseRequestForm(forms.Form):
     shortIdentifier = forms.CharField(label='Short identifier', max_length=25)
     sourceUrl = forms.CharField(label='Source / URL', required=False)
     osiApproved = forms.CharField(label="OSI Status", widget=forms.Select(choices=OSI_CHOICES))
-    notes = forms.CharField(label='Notes', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
+    comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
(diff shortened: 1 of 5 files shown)


Ugtan:
@goneall For now i have kept the `licenseNotes` field in the views to be a empty string as i don't know how the legal team would like to use that field and also, i will wait for your reply for any further changes.

After submitting the license the issue looks like [this](https://github.com/spdx/TEST-LicenseList-XML/issues/8). From the license i submitted i found out that there are no validations checks to check if the user enter a valid url. We should ensure that it is a url and not a regular string right?

goneall:
> We should ensure that it is a url and not a regular string right?

I think that would be a good idea.  If you don't mind adding that to this PR, I can merge that in as well.

Ugtan:
@goneall Sure. I will add that as well.

goneall:
@jlovejoy For the notes and comments, we are using comments as text the license submitter adds when submitting a license.  The comments just go into the github issue and not the XML.

Do you see any need for the license submitter to be allowed to also add notes for the license or is that something that only the legal team should manage?  IMHO, we should only allow the comments and not confuse the user with notes in the license submittal form.

Ugtan:
@goneall Done. Please have a look at it.

goneall:
@Ugtan @jlovejoy Talked to Paul today from the legal team and we both though there is no need to include the license-XML notes in the form.

@Ugtan Let's go ahead and remove the licenseNotes field from the code just to clean it up a bit.  After that, I'll merge and deploy.

Ugtan:
> @Ugtan Let's go ahead and remove the licenseNotes field from the code just to clean it up a bit. After that, I'll merge and deploy.
I agreed with you on not including the notes via forms. But, removing the `licenseNotes` will remove the field notes from the license xml. So, you want to remove them from the license that are submitted as well? @goneall

goneall:
@Ugtan the `<notes>` element is optional in the XML, so we could remove it.   After going back and looking at the code again I'm now thinking it is only a couple lines of code we would be cleaning up.  Leaving it would keep a hook in case we want to add notes in the future.  Let's leave it as is.  I'll merge and deploy.

Thanks for the contribution!

Ugtan:
Okay. I just wanted to make sure that it should be removed from the license xml as well.

And I'm glad i could contribute 😊

Issue #94, opened 2019-03-15
Convert mouse pointer on hover


techytushar:
On the `/app/license_requests/` and `/app/archive_requests/` page, the mouse pointer should change to finger when the user hovers over the table items (license names) so that he knows that the items are clickable. In the current version it feels like they are just for displaying the info and not clickable. There can also be some sort of highlighting on mouse hover.

ithompson4:
Hello @techytushar. Can I work on this problem?

techytushar:
Sure, go ahead :+1:

ithompson4:
Hello @techytushar.
I was able to run the app and populate app_licensenames db table. But when I open `/app/license_requests/` and `/app/archive_requests/ pages`, the tables are empty. How can I populate the database tables? Thanks.

Ugtan:
One and the easy way would be to run the command ``` python manage.py loaddata ./app/testFiles/licenseRequests.json ``` as given [here](https://github.com/spdx/spdx-online-tools/tree/master/src/app/testFiles) or else just create data in database.

ithompson4:
Thanks, it worked.

ithompson4:
Could you please confirm that for the clickable fields under the columns Fullname, Short Identifier, License author name, and Submission Date (UTC) the pointer should be changed to a finger when it hovers over them.
Also, that each of those clickable fields should be highlighted when hovered over.

rtgdk:
@ithompson4 You can also use super admin Django feature to add your own licensename data rows. Also, we want that the mouse pointer should  be changed to `hand` (like when we hover over any link on google) and the hovered data row should be highlighted slight differently that the rest of the rows.

ithompson4:
I also noticed that on the `app/archive_requests/` page when I click on the `Unarchive` button, the “Unarchive” text is underlined in red. 
Is it a potential bug? Or does it work as designed? Can I raise an issue? Thanks.

rtgdk:
@ithompson4 I didn't notice the above bug. Can you please paste a screenshot of it?

ithompson4:
![Screenshot](https://user-images.githubusercontent.com/16367007/54665946-dcca5500-4abe-11e9-8c55-446b499a1af0.png)

rtgdk:
[...]

PR #102, opened 2019-03-20
Convert mouse pointer on hover Issue #94

ithompson4:
Fixed issue #94 Convert mouse pointer on hover:
Added css for `archive-requests.html` and `license-requests.html` files inside base.html
Fixed the issue with the Archive/unarchive button by adding type=`'button'` inside `<input>` tag.

### DIFF
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -31,7 +31,7 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
-              <input id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive' />
+              <button id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive'>Unarchive</button>
             </td>
           </tr>
         {% endfor %}
@@ -43,10 +43,6 @@
 </div>
 {% endblock %}
 {% block script_block %}
-
-<style>
-    th {text-align: center;}
-</style>
 
 <script type="text/javascript">
   $(document).ready(function () {
@@ -70,7 +66,7 @@ $('tbody tr').each(function (i,e) {
       var $licenseId = $(this).closest("tr").attr('id');
       location.href="/app/archive_requests/" + $licenseId;
     });
-    $('body').on('click', 'input[id^="unarchive_button"]', function (en) {
+    $('body').on('click', 'button[id^="unarchive_button"]', function (en) {
       en.stopImmediatePropagation();
       $("#modal-header").addClass("yellow-modal");
 	    $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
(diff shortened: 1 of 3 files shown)


src/app/templates/app/archive_requests.html
```diff
@@ -46,6 +46,9 @@
 
 <style>
     th {text-align: center;}
+    .clickableRow {cursor: pointer;}
+    table#requests_table.dataTable td:hover {background-color: #ddd;}
```
rtgdk: We would like to show the whole row highlighted not a single data field. Use `tr` instead of `td`. Also, don't use the table id to set the style. This would be redundant if we change the name of the table in future. Rather use the style tag directly in the table tag above or set this default in the base.html `style` tag.

src/app/templates/app/license_requests.html
```diff
@@ -46,6 +46,9 @@
 
 <style>
     th {text-align: center;}
+    .clickableRow {cursor: pointer;}
+    table#requests_table.dataTable td:hover {background-color: #ddd;}
+    table#requests_table.dataTable .bt:hover {background-color: #f9f9f9;}
```
rtgdk: Same for these one

rtgdk:
@ithompson4 I suggested some changes above. I just noticed that we have style tag in many different html files in app. If you can move the `<style>` tag from all these html files to the `base.html` <style> tag it would be appreciated. 
We should have only one file containing all the css related tag.

src/app/templates/app/archive_requests.html
ithompson4: Moved `<style>` to base.html, used css for `tr` tag without id.

src/app/templates/app/license_requests.html
ithompson4: Moved `<style>` to base.html, used css for `tr` tag without id.

src/app/templates/app/archive_requests.html
```diff
@@ -31,7 +31,7 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
-              <input id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive' />
+              <input type='button' id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive' />
```
rtgdk: We can change it to `<button>` tag instead `<input>`

ithompson4: Replaced with `<button>` tag

rtgdk:
@ithompson4 You can also combine the `style` tag from `xml_upload` and `license_information` to `base.html`.  I just checked that license_information uses specific ids to modify style.Can you generalise the style by adding a separate class for that? If we can use simple bootstrap tags instead of fixing the margin and font to some pixels, that would be great. Also, squash your commits into single one.

ithompson4:
@techytushar, I will squash my commits. 
@rtgdk, I will do it. Could it be better to create a new pull request for updates to the `xml_upload` and `license_information` files to keep better track of their changes? 
Thanks.

rtgdk:
@ithompson4 Cool, create a new PR for that.

ithompson4:
OK. Once you merge my commit I get the latest updates and continue working on it. Thanks.

rtgdk:
@ithompson4  you can work on the master branch. I will merge this with the new PR you will create and then merge the new pr with the master.

rtgdk:
@ithompson4 Changes looks good. Thanks! Merging.

PR #104, opened 2019-03-21
Add support for documenting API

### LINKED ISSUE #91: Add support for documenting API
Although, the wiki [here](https://github.com/spdx/spdx-online-tools/wiki/REST-API-Fields-Request-and-Response) is enough to interact with the API but adding a support for [API documentation](https://www.django-rest-framework.org/topics/documenting-your-api) will help people new to the repo, understand what the API can do without access to source code. It will also help them to see all the API endpoints available so that they can use these endpoints to test and analyze the requests and responses of the API endpoints which is not really possible in entering curl commands.

Ugtan:
Fixes #91

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ lxml==4.2.5
 social-auth-app-django==2.1.0
 #for testing
 selenium==3.14.1
+drf-yasg==1.14.0
--- a/src/src/settings.py
+++ b/src/src/settings.py
@@ -50,6 +50,7 @@ INSTALLED_APPS = [
     'api',
     'rest_framework',
     'social_django',
+    'drf_yasg',
 ]
 
 MIDDLEWARE = [
--- a/src/src/urls.py
+++ b/src/src/urls.py
@@ -33,7 +33,20 @@ from django.conf import settings
 from django.conf.urls.static import static
 from rest_framework import routers
 from api import views
+from rest_framework import permissions
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
 
+schema_view = get_schema_view(
+   openapi.Info(
+      title="SPDX Online Tools API",
+      default_version='v1',
+      description="API Documentation for SPDX Online Tools used to understand the capabilities of the \
+          tools by testing the requests and responses of the various API endpoints."
+   ),
+   public=True,
+   permission_classes=(permissions.AllowAny,),
+)
 
 router = routers.DefaultRouter()
 #router.register(r'users', views.UserViewSet)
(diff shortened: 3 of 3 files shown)


Ugtan:
@rtgdk I just found out why travis is failing. Drf-yasg works with django rest framework version 3.7.7 or greater which i didn't notice as when i install drf-yasg in my local machine it automatically upgraded the drf to 3.7.7. Moreover, this is also not working well with the current source, so, i think it is better to just drop this idea and use Postman instead. What do you think?

rtgdk:
@Ugtan Cool, go for postman.

Ugtan:
@rtgdk I was talking about this [tool](https://www.getpostman.com/downloads/) in which there is no need to enter curl commands just running the server and using the various api endpoints work well, i have tested it too. So, Should i just remove my above commit and add about this tool in the readme section?

rtgdk:
@Ugtan Yeah I know about postman. I would like if the documentation can be created by postman with as little changes as possible to the api structure and source code.

Ugtan:
There is no need to change anything, i have tried using one of the api endpoints and it is working fine. I will try using the other api endpoints as well and will update this PR soon!

Ugtan:
@rtgdk Please have a look on the documentation [here](https://documenter.getpostman.com/view/5806333/S17rvU4c#intro). I would like to have your thoughts on this. Thanks.

rtgdk:
@Ugtan  Looks great. I like the different languages api helper functions. Although, installing postman to run these seems not a good idea since its a third party tool.(If there's a way to run/test these endpoints from the online-tool itself, that would be great, I think swagger docs provide that ability, can you check?). Is there a way to port these documentation of different languages provided by postman into the online-tools on a separate url `/documentation` or port these to github wiki pages since using the [above url](https://documenter.getpostman.com/view/5806333/S17rvU4c) is not efficient.

Ugtan:
> If there's a way to run/test these endpoints from the online-tool itself, that would be great, I think swagger docs provide that ability, can you check?

I have already tried using the swagger docs, the problem i am facing with swagger and coreapi is that they are not providing options to upload file with the current source code. I tried looking around but not able to find a solution.



> Is there a way to port these documentation of different languages provided by postman into the online-tools on a separate url `/documentation` or port these to github wiki pages since using the [above url](https://documenter.getpostman.com/view/5806333/S17rvU4c) is not efficient.

Postman pro allows to serve the documentation on  a custom domain but the main disadvantage is its paid. I agree with you that using the above url is not efficient. I will try to figure something out.

rtgdk:
@Ugtan @goneall As mentioned on https://github.com/spdx/spdx-online-tools/pull/323, not sure if the generated swagger specs are correct or are we missing anything, closing this one.

PR #114, opened 2019-04-16
Add instructions for form fields in license submit

### LINKED ISSUE #106: license submit: instructions for fields/form
is it possible to add some instructions at the top of the page?

similarly, could we add hints for how to fill out some of the fields - maybe via a popup or via some text in the field that disappears when you then start typing?

Ugtan:
Fixes #106

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -23,7 +23,9 @@
 
     <div class = "form-group">
       <div class="col-sm-12">
-      <label class="control-label col-sm-3" >License's Full Name</label>
+      <label class="control-label col-sm-3" >License's Full Name
+        <span class="glyphicon glyphicon-question-sign" id="fullnameInfo" data-toggle="popover" data-placement="auto left"></span>
+      </label>
         <div class="col-sm-4">
           {{ form.fullname }}
         </div>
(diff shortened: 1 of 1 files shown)


rtgdk:
The popover just flicker on small screen. Try to reduce your browser's screen and hover over the icon. (it might just be my browser, not sure) I am also not sure if it works correctly on mobile.

Still using bootstrap model is much better than simple hover. But I am not sure if this much text should be shown in hover. I think it should be condensed and link to https://spdx.org/spdx-license-list/license-list-overview#fields should be provided for further reference.

goneall:
@rtgdk Thanks for checking that.

I just emulated a phone using Chrome and the popover text doesn't show up when the screen is too small.  It looks like we need to go with reduced text and a link to make this work on a mobile phone.

I think most people will be using a desktop, but we should make it work for phone as well.

Ugtan:
@rtgdk @goneall I have adding a patch for the popover considering your feedback for making it more mobile friendly. The patch seems to work fine. Although, still some of the text may not be shown for the `shortIdentifier` as it is too long and will be exceeding the screen size and user has to scroll a bit to be able to see whole info. I will try to fix this, other than that i think it works well for the phones as well.

Ugtan:
> I just emulated a phone using Chrome and the popover text doesn't show up when the screen is too small

That is probably due to the fact that hover doesn't exist for touch screen devices(except iphone). But i have fixed it and the popover is triggered on click in case of mobile phones and on hover in case of desktops. I have also tested it(on firefox) and it is working. But it doesn't show up as it should on mobile but its not too bad as well. I feel like a bit less text would make it better(if possible).

goneall:
@Ugtan I just check the latest with your patch and the mobile works fine now.

rtgdk:
LGTM. Still not too sure about the length of the hover text or if we should add a link to https://spdx.org/spdx-license-list/license-list-overview#fields . @goneall @techytushar  If it looks good to you, go ahead and merge.

goneall:
I'll go ahead and merge and deploy so that @jlovejoy can review.

PR #119, opened 2019-05-17
Add validations for license submittal fields

### LINKED ISSUE #115: Validate license submittal fields
This license name and license ID can be validated using the rules documented in the pop over text.  We can pick a max. length of a short identifier - perhaps 25 characters.  We can make sure all characters are valid.

For the license name, we can check for starting with "the", parenthesis, etc.

Check the pop over text for what can be validated.

This is related to issue #106

Ugtan:
Fixes #115

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -188,17 +188,18 @@ $(document).on('click', '#addSourceUrl', function(){
 <script type="text/javascript">
 function checkform() {
   $(".form-control").removeAttr('style');
-  if ($('#fullname').val().trim() === "") {
+  if (validate_fullname()) {
     scrollUpAndHighlight($('#fullname'))
-    return ("Please enter the license fullname.");
+    message = validate_fullname()
+    return (message + "Hover over to '?' for more info.");
   }
-  else if (($('#shortIdentifier').val().trim() == "")){
+  else if (($('#shortIdentifier').val().trim() == "") || !validate_identifier()){
     scrollUpAndHighlight($('#shortIdentifier'))
-    return("Please enter the identifier.");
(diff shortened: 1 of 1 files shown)


Ugtan:
I have added the following validation checks for the listed license submittal fields:
1. **License Fullname**
The validation checks added for the license fullname are-
[x] Restrict words starting with *the* for alphabetical sorting purposes.
[x] No commas in the license fullname or exception.
[x] Word *version* shouldn't be spelled out instead use 'v'.
[x] No period or space between version 'v' and version number.

2. **Short Identifier**
[x] Short identifiers must consists of ASCII letters (A-Za-z), digits (0-9), full stops (.) and hyphen or minus signs (-) only
P.s Short identifier cannot be more than 25 characters so, i don't think there is any need to validate it.

Please have a look and let me know if  i left some checks that needs to be covered. @goneall @techytushar

Ugtan:
> you could add text to describe what the valid text would be.

Did you mean to add more description of the error or just give example of what valid field would look like? I am not able to understand, can you be more descriptive?

goneall:
> Did you mean to add more description of the error or just give example of what valid field would look like? I am not able to understand, can you be more descriptive?

An ideal solution would be something very specific like "an invalid character '%' was found in the license ID".  Also, a very acceptable solution would be just to describe what is expected like "The license ID is invalid, short identifiers consist of ASCII letters ..."

I'll leave it to you how specific since there is a bit more work involved in giving the detailed descriptions.

Ugtan:
> An ideal solution would be something very specific like "an invalid character '%' was found in the license ID". Also, a very acceptable solution would be just to describe what is expected like "The license ID is invalid, short identifiers consist of ASCII letters ..."

Okay, got it.


> I'll leave it to you how specific since there is a bit more work involved in giving the detailed descriptions.

Yeah, I will figure something out.

Ugtan:
@goneall I have added a patch considering your feedback, please have a look. Thank you

goneall:
@Ugtan Changes look good.

I'll go ahead and merge.

PR #127, opened 2019-07-17
Compare submitted licenses with the rejected as well as yet not approved licenses.

Ugtan:
To Check if the submitted licenses matches with that of the once that are already rejected or yet not approved licenses by the legal team.

### DIFF
--- a/src/app/utils.py
+++ b/src/app/utils.py
@@ -11,14 +11,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import requests
-import json
 import base64
+import json
 import logging
-from app.models import UserID, User
+import re
 import socket
-from django.conf import settings
 
+import requests
+from django.conf import settings
+from spdx_license_matcher.computation import get_close_matches
+
+from app.models import User, UserID
+
+from .models import LicenseRequest
 
 NORMAL = "normal"
 TESTS = "tests"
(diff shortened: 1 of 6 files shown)


Ugtan:
@goneall @techytushar I have tested it and it is working on my machine I will attach the SC but still I would encourage to test it, just a note for the development it will only work for the issues you have created as the license request you create will be present on the local database but it won't be an issue in the production. Here's the SC
Screenshot_20190720_022052

goneall:
Looks good to me.

@techytushar Let me if you think this is OK to merge.

src/app/utils.py
```diff
@@ -285,3 +291,184 @@ def createLicenseNamespaceIssue(licenseNamespace, token, urlType):
...
+    """
+    licenseIds = []
+    licenseTexts = []
+    for issue in issues:
+        if issue.get('pull_request') is None:
+            licenseInfo = issue.get('body')
+            if '[SPDX-Online-Tools]' in issue.get('title'):
+                licenseIdentifier = re.search(r'(?im)short identifier:\s([a-zA-Z0-9|.|-]+)', licenseInfo).group(1)
+                licenseIds.append(licenseIdentifier)
+                licenseXml = str(LicenseRequest.objects.get(shortIdentifier=licenseIdentifier).xml)
```
techytushar: There should be exception handling here, as its giving me Query does not error .

Ugtan: That error is probably due to no issues being labeled as new request or reject ones but good point I will add error handling to it soon. Thanks for pointing that out. :)

src/app/utils.py
```diff
@@ -285,3 +291,184 @@ def createLicenseNamespaceIssue(licenseNamespace, token, urlType):
...
+                textStr = textStr[:-7].strip().replace('&lt;', '<').replace('&gt;', '>').strip()
+            data['text'] = textStr.strip()
+        else:
+            data['text'] = ''
+    except Exception as e:
+        data['text'] = ''
+    return data
+
+
+def get_xml():
```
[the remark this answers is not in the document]
Ugtan: @techytushar So, apparently the functions(`get_xml` and `download_xml_file`) were doing nothing and wasn't even used anywhere else in the code except within themselves. I just realised now and I have removed them. These were added in this [commit](https://github.com/spdx/spdx-online-tools/pull/31/commits/31c0ed4b437a3239ab18da4645ee1313156c4d3f).

Ugtan:
@techytushar @goneall I have updated the branch with the requested changes. Please have a look. Thanks

goneall:
Great - I'll go ahead and merge

PR #130, opened 2019-08-09
Fix xvfb-run errors in Travis script

goneall:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,6 @@ install:
 script:
   - sudo systemctl status xvfb
   - cd src
-  - xvfb-run python manage.py makemigrations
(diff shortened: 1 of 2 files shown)


goneall:
I attempted to fix the travis build failures.

Adding a -a to the xvfb-run commands seems to resolve the problem (solution referenced in https://stackoverflow.com/questions/16726227/xvfb-failed-start-error).

However, now I am getting the following error on the make-migrations:

`ImportError: numpy.core.multiarray failed to import`

@rtgdk I don't think this is related to my fix.  If you agree, go ahead and merge and we can add a different issue for the import error.

rtgdk:
@goneall @techytushar Travis build working fine after updating requirements. Merging.

Issue #144, opened 2020-02-08
Running the server shows an error no module named secret in settings.py file


anshuldutt21:
While running the command python manage.py runserver an error comes up showing no module exists named secret in settings.py file. It should be renamed as
from .secret import getgithubkey and others,  since we are accessing the file in the same directory as of the settings.py file

rtgdk:
@anshuldutt21 Are you sure this is not specific to the python version/ virtualenv setup / Django version? because it has beem working fine for us in the past.

Did you setup your local repo using the instruction in README?

anshuldutt21:
@rtgdk I think you are right about this issue being specific to the python version. I tried resetting up and it worked.

rtgdk:
Closing this then.

PR #145, opened 2020-02-08
change settings file import error issue#144

anshuldutt21:
Change line from secret to .secret.
Fix for issue #144.


rtgdk:
I am not sure if we need his change, since this might be the affect of a specfic python/django version.

Issue #148, opened 2020-02-14
Submit new license request doesnt work since License Request is not imported


anshuldutt21:
We need to import the License Request from the models to make it work

rtgdk:
Can you post the error? @anshuldutt21

anshuldutt21:
![Screenshot from 2020-02-15 21-46-29](https://user-images.githubusercontent.com/45336622/74591301-d4ff9180-503c-11ea-86c8-cd1f90332a18.png)
This is the error that appears when we submit a new license request.

rtgdk:
@anshuldutt21 But it works in production, can you please check your local setup/ virtualenv? I think this is a python import error specific to the version.

Also, did you try and replicate this in production? https://spdxtools.sourceauditor.com/

anshuldutt21:
@rtgdk It works in production. But I dont understand how it is working if we have'nt imported the model License Request in utils.py. The name will remain undefined. Also I  think there won't be any harm in adding a word along with from app.models import User, UserId. It will make it less prone to version specific errors.

rtgdk:
@anshuldutt21 If you think it's a real issue, you can raise a PR and we will check that on our end.

techytushar:
Closing this, resolved in #149

PR #149, opened 2020-02-17
Import LicenseRequest and namespace to utils

anshuldutt21:
Import License Request from app.models to overcome the error raised in form submission - 'License Request does not exist', line 440 in utils.py.
Issue #148

### DIFF
--- a/src/app/utils.py
+++ b/src/app/utils.py
@@ -28,7 +28,7 @@ from spdx_license_matcher.computation import (checkTextStandardLicense,
 from spdx_license_matcher.difference import get_similarity_percent
 from spdx_license_matcher.utils import get_spdx_license_text
 
-from app.models import User, UserID
+from app.models import User, UserID, LicenseRequest, LicenseNamespace
 
 NORMAL = "normal"
 TESTS = "tests"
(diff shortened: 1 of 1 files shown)


rtgdk:
Change makes sense. I think production does not have updated spdx-online-tools that's why it was working in production.

@goneall When was the last time production got updated with master?

src/app/utils.py
```diff
@@ -28,7 +28,7 @@
 from spdx_license_matcher.difference import get_similarity_percent
 from spdx_license_matcher.utils import get_spdx_license_text
 
-from app.models import User, UserID
+from app.models import User, UserID, LicenseRequest, LicenseNamespace
```
rtgdk: Instead of LicenseRequest, it should be License on line 437 in utils and everywhere else in utils. `shortIdentifier` is a field of License not LicenseRequest

anshuldutt21:
@rtgdk Is this change fine? I can edit the pull request message.

rtgdk:
@anshuldutt21  Thanks! LGTM. Merging.
I am not sure how we didn't catch this issue in testing.

Issue #93, opened 2019-03-12
Add versioning information to the application


goneall:
Since we are now deploying multiple versions of the app, we should add a version to the application and display the version in the about page. It would also be nice to display the version of the SPDX tools used in the backend as well.

rtgdk:
Totally agree with you. We should also keep track of the major changes done during the development of a version (both java and online-tools).
Ideas I can think of right now:
1. Using `settings.py` file for maintaining the current version
2. Using git tag to mark commits to a specific version and use GitHub releases feature.
3. Having a wiki where we maintain changes done in a version.
I am open to other ideas as well of tagging and maintaining version and displaying the changes in some other/third party way, if its more convenient.

techytushar:
We can also use Milestone on GitHub to set feature requests for further versions.

abhishekspeer:
@goneall @rtgdk @techytushar Can I work on this issue?

rtgdk:
@abhishekspeer  Sure go ahead!

Ugtan:
@abhishekspeer Are you still working on this issue? if not, i would like to work on it.

abhishekspeer:
Thanks for asking @Ugtan 
I had my University exams earlier. I’m working on it right now and will submit a PR soon. Thanks again. :)

Ugtan:
No issues 😊 @abhishekspeer

rtgdk:
We would like to include the changelog in the wiki. (Format of the changelog should also be specified in that wiki). Before that, you should present any versioning strategy that would be best for this project. We can keep the online-tools version and java-tools version in settings.py or any other file and show it at the bottom of the web page. We should also create a wiki on how/when to change a version. Some application follow <Major.Minor.Patch> strategy. When a patch is added we change only the last digit. For eg: `1.0.1` to `1.0.2`. Also, if we can use the github release feature that would be great. Basically you can describe your strategy on the issue itself or include some PR to show how you plan to implement it. Once we have that, the mentors will help you guide what extra things can be included.

anshuldutt21:
@rtgdk its best to include a file (versions) to show the java tools version and display the value in the web home page.

rtgdk:
@anshuldutt21 You are welcome to open a PR for the same.

PR #150, opened 2020-02-17
Add version information and render in front page

anshuldutt21:
Add a new file version information to keep record of the versions.
Render the version information in the about page of spdx.
Fixes for issue #93


rtgdk:
@anshuldutt21 LGTM. Although, the version number for java is not correct, I think it is 2.1.8.
Can you also post a screenshot of the change, so that it is easier and quicker for people to review the changes?

anshuldutt21:
@rtgdk Can you tell me where can I find the version of java tools and spdx? I can also write about it in the comments of the version file.

rtgdk:
@anshuldutt21  Even, I found out about the version number from the last commit to that tar. But, you can play around with the tar using java : check https://github.com/spdx/tools

anshuldutt21:
@rtgdk okay. Also, maybe we can start tagging spdx-online-tools and release the version starting from 1.0.

src/src/version.py
```diff
@@ -0,0 +1,11 @@
...
+Version file for displaying the respective version of 
+the tools and online-tools repositories.
+"""
+
+spdx_version = '1.0'
+
+"""
+Visit https://github.com/spdx/tools/releases to know about the tools releases.
+"""
+java_tools_version = '2.1.20'
```
rtgdk: The current tar is not updated with the latest spdx tools release. I think it is 2.1.8 instead (check the last commit on the tar)

anshuldutt21:
@rtgdk Is it fine now? Please review.

src/app/views.py
```diff
@@ -78,7 +80,10 @@ def about(request):
     """ View for about
     returns about.html template
     """
-    context_dict={}
+    context_dict={
+                'spdx_version':spdx_version,
```
rtgdk: Change this to `spdx_online_tools_version`

rtgdk: spdx_version may be confused with with spdx specification version

rtgdk:
@anshuldutt21 LGTM. I have suggested one more change. Need approval from @goneall @techytushar

src/app/views.py
goneall: Agree with changing to online tools version

anshuldutt21:
Please review @rtgdk

anshuldutt21:
Please review @rtgdk

goneall (approved):

Issue #147, opened 2020-02-09
Add command python manage.py makemigrations in the documentation


anshuldutt21:
We should add the command python src/manage.py makemigrations in the docs since the migrations for app doesnt work unless we do makemigrations for it.

rtgdk:
@anshuldutt21  Are you sure about this? Because migrations are already added in the migrations folder. If you still had to run makemigrations look at the diff when the command is run, if it creates some new file please open a PR for the same. 
Also, another way is to remove the folder and then run the command, this will create only one file `0001_initial.py` and then you can find the diff.

The idea is to minimize the no. of commands user needs to run for setting up the project.

anshuldutt21:
@rtgdk yes it does create a new file in the migrations after runninig the command.

rtgdk:
@anshuldutt21 Can you paste the content of it?

anshuldutt21:
# -*- coding: utf-8 -*-
# Generated by Django 1.11.28 on 2020-02-15 16:52
from __future__ import unicode_literals

from django.db import migrations, models
import django.db.models.deletion


class Migration(migrations.Migration):

    dependencies = [
        ('app', '0001_initial'),
    ]

    operations = [
        migrations.CreateModel(
            name='LicenseNamespace',
            fields=[
                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                ('licenseAuthorName', models.CharField(blank=True, default='', max_length=100, null=True)),
                ('fullname', models.CharField(max_length=70)),
                ('shortIdentifier', models.CharField(max_length=25)),
                ('submissionDatetime', models.DateTimeField(auto_now_add=True)),
                ('userEmail', models.EmailField(max_length=35)),
                ('notes', models.CharField(default='', max_length=255)),
                ('xml', models.TextField()),
                ('archive', models.BooleanField(default=False)),
                ('publiclyShared', models.BooleanField(default=True)),
                ('description', models.TextField()),
                ('namespace', models.CharField(max_length=200)),
                ('url', models.CharField(max_length=200)),
                ('license_list_url', models.URLField(max_length=250)),
                ('github_repo_url', models.URLField(max_length=250)),
                ('promoted', models.BooleanField(default=False)),
            ],
            options={
                'verbose_name': 'LicenseNamespace',
                'verbose_name_plural': 'LicenseNamespaces',
            },
        ),
        migrations.CreateModel(
            name='OrganisationName',
            fields=[
[...]

PR #152, opened 2020-02-18
Make a migrations file to avoid using the command makemigrations

anshuldutt21:
Make a migrations file in the app folder to avoid using makemigrations in the READme file.
Fixes for issue #147


rtgdk:
@anshuldutt21  LGTM. Merging. Thanks!

PR #154, opened 2020-03-29
Solve archive requests issue #107

### LINKED ISSUE #107: Archive button only visible to logged in users
Currently the Archive button on the `license_requests/` page and Unarchive button on the `archive_requests/` page is visible to everyone, so anyone can just press the buttons and mess things up. These buttons should only be visible to the users who are logged in. What do you think @goneall

anshuldutt21:
The users will now need to login in to archive or unarchive the request.
Only, the repo collaborators will be able to archive the request.
Fixes #107

### DIFF
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -32,7 +32,11 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
+              {% if github_login %}
+              {% if authorized %}
               <button id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive'>Unarchive</button>
+              {% endif %}
+              {% endif %}
             </td>
           </tr>
         {% endfor %}
@@ -92,15 +96,6 @@ $('tbody tr').each(function (i,e) {
                 show: true,
           })
        }
-      }
-      else {
-        $("#modal-header").addClass("yellow-modal");
-        $("#modal-body").html('You must login to use this functionality');
-        $("#myModal").modal({
-                backdrop: 'static',
-                keyboard: true,
-                show: true,
-          })
       }
 
       var license = $(this).closest('tr');
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -32,10 +32,8 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
-              {% if github_login %}
-              {% if authorized %}
+              {% if github_login and authorized %}
               <button id='unarchive_button{{elem.id}}' class='btn btn-success' width='25px' value='Unarchive'>Unarchive</button>
-              {% endif %}
               {% endif %}
             </td>
           </tr>
@@ -75,28 +73,14 @@ $('tbody tr').each(function (i,e) {
     });
     $('body').on('click', 'button[id^="unarchive_button"]', function (en) {
       en.stopImmediatePropagation();
-      var githubLogin =  "{{ github_login }}" ;
-      if ((githubLogin) != "None"){
-        if (" {{ authorized }} ") {
-          $("#modal-header").addClass("yellow-modal");
-          $("#modal-body").html('You are not permitted to unarchive this license.');
-          $("#myModal").modal({
+       $("#modal-header").addClass("yellow-modal");
+	     $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
+       $("#modal-body").html('Are you sure you want to unarchive this license?');
+       $("#myModal").modal({
                 backdrop: 'static',
                 keyboard: true,
                 show: true,
           })
-        }
-        else{
-              $("#modal-header").addClass("yellow-modal");
-	            $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
-              $("#modal-body").html('Are you sure you want to unarchive this license?');
-              $("#myModal").modal({
-                backdrop: 'static',
-                keyboard: true,
-                show: true,
-          })
-       }
-      }
 
       var license = $(this).closest('tr');
       $(document).on('click','button#confirm_unarchive', function(event){
@@ -105,7 +89,6 @@ $('tbody tr').each(function (i,e) {
         unarchiveLicense(license);
       });
     });
-  });
 
 function unarchiveLicense(license) {
   var archive = 'False';
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -18,7 +18,9 @@
           <th>Short Identifier</th>
           <th>License author name</th>
           <th>Submission Date (UTC)</th>
+          {% if github_login and authorized %}
           <th>Unarchive</th>
+          {% endif %}
         </tr>
       </thead>
       <tbody>
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -7,7 +7,6 @@
 
 <div id="messages" class="messages">
 </div>
-
 <p class ="lead"> {{ error }}</p>
 <div class="panel panel-default">
 <div class="panel-heading"> <p class="lead">Archive License Requests List</p> </div>
@@ -73,15 +72,14 @@ $('tbody tr').each(function (i,e) {
     });
     $('body').on('click', 'button[id^="unarchive_button"]', function (en) {
       en.stopImmediatePropagation();
-       $("#modal-header").addClass("yellow-modal");
-	     $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
-       $("#modal-body").html('Are you sure you want to unarchive this license?');
-       $("#myModal").modal({
-                backdrop: 'static',
-                keyboard: true,
-                show: true,
-          })
-
+      $("#modal-header").addClass("yellow-modal");
+	    $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
+      $("#modal-body").html('Are you sure you want to unarchive this license?');
+      $("#myModal").modal({
+        backdrop: 'static',
+        keyboard: true,
+        show: true,
+      })
       var license = $(this).closest('tr');
       $(document).on('click','button#confirm_unarchive', function(event){
         event.preventDefault();
@@ -89,6 +87,7 @@ $('tbody tr').each(function (i,e) {
         unarchiveLicense(license);
       });
     });
+  });
 
 function unarchiveLicense(license) {
   var archive = 'False';
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -7,6 +7,17 @@
 
 <div id="messages" class="messages">
 </div>
+{% if github_login  %}
+<p id="githubLogin" hidden>True</p>
+{% else %}
+<p id ="githubLogin" hidden> False </p>
+{% endif %}
+{% if authorized  %}
+<p id="authorized" hidden>True</p>
+{% else %}
+<p id ="authorized" hidden> False </p>
+{% endif %}
+
 <p class ="lead"> {{ error }}</p>
 <div class="panel panel-default">
 <div class="panel-heading"> <p class="lead">Archive License Requests List</p> </div>
@@ -70,14 +81,35 @@ $('tbody tr').each(function (i,e) {
     });
     $('body').on('click', 'button[id^="unarchive_button"]', function (en) {
       en.stopImmediatePropagation();
-      $("#modal-header").addClass("yellow-modal");
-	    $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
-      $("#modal-body").html('Are you sure you want to unarchive this license?');
-      $("#myModal").modal({
-        backdrop: 'static',
-        keyboard: true,
-        show: true,
-      })
+      var githubLogin = $("#githubLogin").text();
+      var authorized = $("#authorized").text();
+      console.log(authorized);
+      if (githubLogin == "True"){
+        if (authorized) {
+          $("#modal-header").addClass("yellow-modal");
+          $("#modal-body").html('You are not permitted to unarchive this license.');
+          $("#myModal").modal({
+                backdrop: 'static',
+                keyboard: true,
+                show: true,
+          })
+        }
+        else{
+              $("#modal-header").addClass("yellow-modal");
+	            $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
+              $("#modal-body").html('Are you sure you want to unarchive this license?');
+              $("#myModal").modal({
+                backdrop: 'static',
+                keyboard: true,
+                show: true,
+          })
+       };
+      }
+      else {
+        $("#modal-header").addClass("yellow-modal");
+        $("#modal-body").html('You must login to use this functionality');
+      }
+
       var license = $(this).closest('tr');
       $(document).on('click','button#confirm_unarchive', function(event){
         event.preventDefault();
--- a/src/app/templates/app/archive_requests.html
+++ b/src/app/templates/app/archive_requests.html
@@ -7,16 +7,6 @@
 
 <div id="messages" class="messages">
 </div>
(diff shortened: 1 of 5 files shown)


anshuldutt21:
@rtgdk Can you please have a look at this?

src/app/templates/app/archive_requests.html
```diff
@@ -7,6 +7,17 @@
...
+<p id="githubLogin" hidden>True</p>
+{% else %}
+<p id ="githubLogin" hidden> False </p>
+{% endif %}
+{% if authorized  %}
+<p id="authorized" hidden>True</p>
+{% else %}
+<p id ="authorized" hidden> False </p>
+{% endif %}
+
```
rtgdk: Totally inefficient and insecure way of doing this.
Think about it this way, assume I have access to "Inspect Element" function in my browser. If I change this html tag value from False to True, will I be able to access the archive request feature?

rtgdk:
@anshuldutt21 Take a look at this base template file : https://github.com/spdx/spdx-online-tools/blob/8e2974f73a412d3c576f874379d5f19adc06bd2f/src/app/templates/app/base.html#L162 for inspiration on how to do this.

src/app/templates/app/archive_requests.html
```diff
@@ -70,14 +71,38 @@
...
-      $("#modal-header").addClass("yellow-modal");
-	    $(".modal-footer").html('<button class="btn btn-success" id="confirm_unarchive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
-      $("#modal-body").html('Are you sure you want to unarchive this license?');
-      $("#myModal").modal({
-        backdrop: 'static',
-        keyboard: true,
-        show: true,
-      })
+      var githubLogin =  "{{ github_login }}" ;
+      if ((githubLogin) != "None"){
```
rtgdk: This again has security flaws. User can see this js in the source page for html. Now they can run the same thing on console to add these html elements bypassing the if else check with simple `gitHubLogin = true`. Please see this : https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#if
to introduce this check in html itself.

anshuldutt21: @rtgdk  The function runs whenever the user clicks on the POST button. Without using java script, How will I be able to write the function. or I could instead of declaring the variable just write if ( "{{ github_login }}" != "None"?

rtgdk: @anshuldutt21 Have you had a look at https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#if

We can keep the if..else in the html itself. 
Have a look at line: https://github.com/spdx/spdx-online-tools/blob/8e2974f73a412d3c576f874379d5f19adc06bd2f/src/app/templates/app/base.html#L162

You can find the same if..else in any of my django projects.

anshuldutt21: Hi @rtgdk , I apologize for replying too late.
`
	var githubLogin = $("#githubLogin").text();
	/* if user not authenticated using GitHub, display modal with login button */
	if(githubLogin == "False"){
		$("#modal-header").removeClass("red-modal green-modal");
		$("#modal-header").addClass("yellow-modal");
		$(".modal-footer").html('<button class="btn btn-success" id="github_auth_begin"><span class="glyphicon glyphicon-ok"></span> Confirm</button>');
		$("#modal-body").html('To submit a license, you must log in using Github.  You will now be redirected to the github login.  Please allow the requested permissions.  If you do not have a github account, you can <a href="https://github.com/">sign up</a> for free or you can email your new license request to <a href="mailto:spdx-legal@spdx.org">spdx-legal@spdx.org</a>.');
        $("#myModal").modal({
           backdrop: 'static',
           keyboard: true,
           show: true
        });
	}
	}
`
[https://github.com/spdx/spdx-online-tools/blob/master/src/app/templates/app/submit_new_license.html](url)
In the code of submit new license as above, it was done in a similar way because by HTML method we could'nt call the function. The same issue is with the license request.html code. Here we are calling the onclick method of button (archive) and then accordingly we are formatting the html of modal using the add function. 
Also, even if a person is able to edit the javascript he wont be able to archive the request since I have added a check in the backend view that only an authorised person can archive a  request.

rtgdk: @anshuldutt21 But the question is, why are we showing the button when user is not logged in or does not have permission - just to show them error dialog box saying "You are not allowed to do this.". Instead, to improve UX we should show the button only when the user is logged in and has perms to do this. This can be done in html if else itself with some backend processing of user info. This would also be easier in terms of the amount of code to be written and will be easily manageable. 
Also, if you find such code pre written like in `submit_new_license.html` feel free to change that too in this PR.

anshuldutt21:
I have done the required changes in the license and the archive requests file. please review it and then I will do the changes in the submit license file as well.

src/app/templates/app/archive_requests.html
```diff
@@ -31,7 +32,11 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
+              {% if github_login %}
+              {% if authorized %}
```
rtgdk, line 36: ifs can be combined.

src/app/templates/app/license_requests.html
```diff
@@ -31,15 +33,16 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
+              {% if github_login %}
```
rtgdk, line 36: no {{authorized}} here

src/app/templates/app/license_requests.html
```diff
@@ -70,14 +73,29 @@
...
+              $("#modal-header").addClass("yellow-modal");
+	            $(".modal-footer").html('<button class="btn btn-success" id="confirm_archive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
+              $("#modal-body").html('You are not permitted to archive this request.');
+              $("#myModal").modal({
+                backdrop: 'static',
+                keyboard: true,
+                show: true,
+          })
+       }
+      }
```
rtgdk, line 97: Can this check be moved in backend?
On Ui we don't show the button for user who are not logged in and does not permission to archive. So, this check on frontend is useless. Also, if a user gains access to send a request to archive , we will check the authenticity in the backend itself and send error code and error messsage.

src/app/templates/app/license_requests.html
```diff
@@ -70,14 +73,29 @@
...
-	    $(".modal-footer").html('<button class="btn btn-success" id="confirm_archive"><span class="glyphicon glyphicon-off"></span> Confirm</button>');
-      $("#modal-body").html('Are you sure you want to archive this license?');
-      $("#myModal").modal({
-        backdrop: 'static',
-        keyboard: true,
-        show: true,
-      })
+      var githubLogin =  "{{ github_login }}" ;
+      if ((githubLogin) != "None"){
+        if (" {{ authorized }} ") {
```
rtgdk, line 78: Why are there always two checks? if `githubLogin` and `authorized` ? can this be combined?

anshuldutt21: basically github login check is for whether the person is logged in by github or not. A person not logged in by github cant be checked for permissions. Authorisation check is for whether the person is a collaborator or not.

src/app/templates/app/license_requests.html
anshuldutt21: Done the changes.

src/app/views.py
```diff
@@ -1290,13 +1290,37 @@ def archiveRequests(request, license_id=None):
...
-        if license_id:
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/tools-python/collaborators/{username}' , headers={'Authorization': 'token {}'.format(token) })
+            if ((test.status_code == 200) or (test.status_code == 204)):
```
rtgdk, line 1302: Add try except here, since this api may result in error. Also, if status code returned is not 200/204, then also we should show some msg to the user.

anshuldutt21: @rtgdk What do you think should be the changes in submit license. In the current version, the  user not logged in gets a message as to he must login in before submitting. If I hide the submit button from the user not logged in, He will not be able to get an idea as to what to do to submit the license.

anshuldutt21: In the backend the user gets a message that permission is denied. but no message should be displayed in the frontend as there is no option(button) for archive.

src/app/views.py
```diff
@@ -1373,13 +1400,40 @@ def licenseRequests(request, license_id=None):
...
+        else:
+            if (request.is_ajax()):
+                ajaxdict = {}
+                ajaxdict["type"] = "auth_error"
+                ajaxdict["data"] = "Please login using GitHub to use this feature."
+                response = dumps(ajaxdict)
+                return HttpResponse(response,status=401)
+            return HttpResponse("Please login using GitHub to use this feature.",status=401)
+    archiveRequests = LicenseRequest.objects.filter(archive='False').order_by('-submissionDatetime')
+    context_dict['archiveRequests'] = archiveRequests
```
rtgdk, line 1430: Should be `licenseRequests`

src/app/views.py
```diff
@@ -1290,13 +1290,40 @@ def archiveRequests(request, license_id=None):
...
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/tools-python/collaborators/{username}' , headers={'Authorization': 'token {}'.format(token) })
+            try:
+                if ((test.status_code == 200) or (test.status_code == 204)):
+                    archive = request.POST.get('archive', True)
```
rtgdk, line 1304: Should be `archive = request.POST.get('archive', False)`

src/app/views.py
```diff
@@ -1290,13 +1290,40 @@ def archiveRequests(request, license_id=None):
...
-        license_id = request.POST.get('license_id', False)
-        if license_id:
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/tools-python/collaborators/{username}' , headers={'Authorization': 'token {}'.format(token) })
```
rtgdk, line 1301: {username} will this work in python in this way?

src/app/views.py
```diff
@@ -1373,13 +1400,40 @@ def licenseRequests(request, license_id=None):
...
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
-    licenseRequests = LicenseRequest.objects.filter(archive='False').order_by('-submissionDatetime')
-    context_dict={'licenseRequests': licenseRequests}
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/spdx-online-tools/collaborators/anshuldutt21' , headers={'Authorization': 'token {}'.format(token) })
```
rtgdk, line 1411: Remove `anshuldutt21`

src/app/templates/app/license_requests.html
```diff
@@ -70,14 +73,14 @@
...
-        show: true,
-      })
+       $("#modal-header").addClass("yellow-modal");
+       $("#modal-body").html('Are you sure you want to archive this license?');
+       $("#myModal").modal({
+                backdrop: 'static',
+                keyboard: true,
+                show: true,
+          })
+
```
rtgdk, line 83: Screenshot 2020-04-28 at 6 38 44 PM
Dialog box does not have submit button.
Please test all things before commiting.

src/app/views.py
rtgdk: Let's leave submit license page then.

src/app/views.py
anshuldutt21: changed it..

src/app/views.py
```diff
@@ -1290,13 +1290,40 @@ def archiveRequests(request, license_id=None):
...
-        license_id = request.POST.get('license_id', False)
-        if license_id:
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/tools-python/collaborators/'+username , headers={'Authorization': 'token {}'.format(token) })
```
rtgdk, line 1301: Going through the code, I found that this would be more suitable if moved in utils.py just like we have function for makePullRequest.

src/app/templates/app/license_requests.html
```diff
@@ -31,7 +31,9 @@
             <td>{{elem.submissionDatetime}}</td>
             <td>
               {% csrf_token %}
+              {% if github_login and authorized %}
               <button id='archive_button{{elem.id}}' class='btn btn-success' width='25px' value='Archive'>Archive</button>
+              {% endif %}
```
rtgdk, line 36: IMO, the "Archive" table header should also be removed if user can't see the archive button. Same with Unarchive.
Screenshot 2020-04-30 at 12 18 35 PM

src/app/views.py
anshuldutt21: Done the changes. I created a function to retrieve the response code. Should I add more functionality to it??

anshuldutt21:
Actually the travis is failing because its not able to locate the id as I have hidden it. In order to work, the tests will have to be modified.

src/app/views.py
```diff
@@ -1290,13 +1290,40 @@ def archiveRequests(request, license_id=None):
...
-        license_id = request.POST.get('license_id', False)
-        if license_id:
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = utils.checkPermission(username,token)
```
rtgdk, line 1301: This whole code for checking user perms should be moved to utils...starting from 1296 line.

src/app/views.py
```diff
@@ -1373,13 +1400,40 @@ def licenseRequests(request, license_id=None):
...
-        license_id = request.POST.get('license_id', False)
-        if license_id:
-            LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
+        if request.user.is_authenticated():       
+            user = request.user
+            """ Getting user info for submitting github issue """
+            github_login = user.social_auth.get(provider='github')
+            token = github_login.extra_data["access_token"]
+            username = github_login.extra_data["login"]
+            test = requests.get('https://api.github.com/repos/spdx/spdx-online-tools/collaborators/'+username , headers={'Authorization': 'token {}'.format(token) })
```
rtgdk, line 1411: utils should be here too.

rtgdk:
@anshuldutt21 Then you would have to update the test (since archive button is visible to some users) and add one more test for permitted github user functionality check.

anshuldutt21:
@rtgdk  I think this PR can be merged now if there are no changes to be done.

src/app/tests.py
```diff
@@ -1119,10 +1120,14 @@ def test_archive_license_requests_feature(self):
         license_name = driver.find_element_by_css_selector('td').text
         self.assertEquals(license_name, "BSD Zero Clause License-00")
         self.assertEquals(LicenseRequest.objects.get(id=license_obj.id).archive, False)
-        driver.find_element_by_id('archive_button' + str(license_obj.id)).click()
-        driver.find_element_by_id('confirm_archive').click()
-        self.assertEquals(LicenseRequest.objects.get(id=license_obj.id).archive, True)
+        if driver.find_element_by_id('archive_button' + str(license_obj.id)):
```
rtgdk, line 1123: When will the test reach here? I think if the user has access to github_login, only then he will be able to see the feature. I think we run the test for a simple user so code will never go into `if`. Can you check for that or is there a way to run the same test with different user? 
You can either create a new issue for that and work on that separately, I'll merge this. Let me know what you find.

rtgdk: @anshuldutt21 Can you create one now, I'll merge this then. Just to be on the safe side of not forgetting about this one. Also, then anyone can pick up the task if you are busy.

anshuldutt21: Done

rtgdk: Thanks! Merging

PR #155, opened 2020-04-07
Popover update

### LINKED ISSUE #151: Innovative popover while submitting license
The current popover in the Submit License page cannot handle scrolling, also it doesn't support mouse click. User has the keep mouse hover over the element to see the info.

It would be more intuitive to have a proper hover over which can open with a click and does not close until another click happens on other part of screen. It should also support scrolling since the text in the info popover is too large.

Any other innovative idea to show these info is welcomed.

ekongobie:
Fixes #151 

Make popover scrollable, displayed on click, and hidden on click anywhere else.


src/app/templates/app/submit_new_license.html
```diff
@@ -170,30 +170,29 @@
 {% block script_block %}
 <script type="text/javascript">
 $(document).ready(function () {
-    var is_touch_device = "ontouchstart" in document.documentElement;
-    $('#fullnameInfo').popover({trigger: is_touch_device ? "click" : "hover", title: "License Full Name", content: "\u2022 The full name may omit certain word, such as 'the,' for alphabetical sorting purposes. \
+    $('#fullnameInfo').popover({trigger: "click", title: "License Full Name", content: "\u2022 The full name may omit certain word, such as 'the,' for alphabetical sorting purposes. \
     <br /> \u2022 No commas in full name of license or exception. \
     <br /> \u2022 The word 'version' is not spelled out; 'v' or nothing is used to indicate license version (for space reasons) \
     <br /> \u2022 For version, use lower case v and no period or space between v and the number. \
     <br /> \u2022 No abbreviations are included (in parenthesis) after the full name.", html: "true", viewport: '.panel-body', container: '.panel-body'});
```
rtgdk, line 177: So much content! Can we save this somewhere like in a file/ javascript variable so it is easy to modify -> for eg if I need to add another point, I have to add `<br />` on every line which is not feasible for non tech people.
Help regarding this would be appreciated.

rtgdk:
@ekongobie Good. i found one more issue in this - When the input box is selected (which is by default) and then we click on the popover button, it doesn't show the pop over. Only when clicked again, the pop over appears. Can you look into it?

Also, I just noticed that we have manually added the content in each popover. Changing this pop over would require extra efforts. Can you look into this and propose a new way to load this content? Let me know if you need more info/help regarding this.

Also, I noticed that we manually

ekongobie:
@rtgdk I am looking into the "click" issue.

The default trigger for the popovers is the "click" action, so I can remove that option completely.
Concerning the popover content, we can definitely read that from a text file.
On it!

ekongobie:
@rtgdk I have solved the popover click issue. However, inorder to solve this, the dismissal of the popover is only done when we click "outside", that is, not on the popover icon. Please test for convenience. I also removed the help text from the javascript and put is in a python file; and then I passed the help text to javascript through form field attributes.

src/app/static/js/editor/codemirror.js
```diff
@@ -9531,10 +9531,10 @@ TextareaInput.prototype.needsContentAttribute = false
 
 function fromTextArea(textarea, options) {
   options = options ? copyObj(options) : {}
-  options.value = textarea.value
-  if (!options.tabindex && textarea.tabIndex)
+  options.value = textarea ? textarea.value : ""
+  if (!options.tabindex && textarea && textarea.tabIndex)
```
rtgdk, line 9535: why do we need this change?

src/app/constants.py
```diff
@@ -0,0 +1,51 @@
+from django.utils.safestring import mark_safe
+
+# License submission form help text
+
+FULL_NAME_HELP_TEXT = mark_safe('\u2022 The full name may omit certain word, such as \'the,\' for alphabetical sorting purposes. <br/>'
```
rtgdk, line 5: Change `\'the,\'` to the

rtgdk: Please review other lines too.

src/app/templates/app/submit_new_license_namespace.html
```diff
@@ -383,4 +397,11 @@
     }
 });
 </script>
+<script type="text/javascript" src="{% static 'js/editor/codemirror.js' %}"></script>
+<script type="text/javascript" src="{% static 'js/editor/script.js' %}"></script>
```
rtgdk, line 401: Why do we need all these scripts? It will increase load time.

rtgdk (changes requested):
@ekongobie Changes look good. Python looks less cluttered than js code, thanks for that! Left some comments before I merge.

src/app/static/js/editor/codemirror.js
ekongobie: @rtgdk I did this to clear up console warning messages displayed when the textarea is null.

src/app/templates/app/submit_new_license_namespace.html
ekongobie: You are right. I did this during testing. I will remove them. Expect an update soon.

src/app/constants.py
ekongobie: Ok

ekongobie: @rtgdk You really do not want single quotes around such words? Please have a look at the image below, you can read and see the use of those quotes. I will remove them if it is still your wish after this.
ek1

rtgdk: @ekongobie My bad, I didn't understand the usage of the backslash and quote.

rtgdk (approved):

rtgdk:
@ekongobie Thanks for working on this!

src/app/templates/app/submit_new_license.html
```diff
@@ -593,6 +590,7 @@
     });
   }
 </script>
+<script type="text/javascript" src="{% static 'js/editor/codemirror.js' %}"></script>
```
rtgdk, line 593: Why do we need this script?

rtgdk (changes requested):

goneall:
@ekongobie If you could resolve the merge conflicts, I'll go ahead and merge this PR.

bact:
Note that

- The Submit new license page no longer has the popover, https://tools.spdx.org/app/submit_new_license/ 
- The popover is still in use in Submit new license namespace page at https://tools.spdx.org/app/submit_new_license_namespace/ -- but this page is not being advertised and not in the menu. May related to these closed issues:
  - https://github.com/spdx/spdx-spec/issues/113
  - https://github.com/spdx/spdx-spec/pull/209
  - https://github.com/spdx/spdx-online-tools/issues/337 - comments confirmed the license namespace is never landed in the spec

bact:
The license namespace features are now all disabled after the merge of https://github.com/spdx/spdx-online-tools/pull/679 . So there is no location that the popover is being used.

rtgdk:
Closing this now based on @bact comments

Issue #58, opened 2018-10-09
Migrate to Django 2.1


rtgdk:
Similar issue #24 

Upgrade the codebase to Python 3 and Django 2.1.
Some work is already done in https://github.com/spdx/spdx-online-tools/tree/python3

SidharajYadav:
How i can contribute
please guide me
I like to solve this problem

autoantwort:
Can be closed. Django 3.2.15 is already used.

Issue #24, opened 2018-03-31
Adding support for Python3


techytushar:
When I try to run the app with `Python 3.5.2` the following error occur:
1. `except jpype.JavaException, ex` is not supported can be changed to `except jpype.JavaException as ex`
2. `from models import` gives `Import error` can be changed to `from api.models`
3. `from urlparse import urljoin` also gives `import error` as there is no module named `urlparse` in Python3 it has been changed to `urllib.parse`
4. Also the `jpype` gives error while installing for python3.6

1 and 2 are fixable but I cannot find a way to fix 3

rtgdk:
Although Django 1.11 is the LTS version, which will be supported for around 3 years. We can create a separate branch to work on Python 3 with Django 2. This would be a good head start.

For 4th, we can use https://pypi.python.org/pypi/JPype1-py3 , although it is technically in development phase. BTW jpype requirements suggest using Python2.7+, so I think it should work fine with python3.6 . What is the exact error you are facing? You can look through the original repo to find your issue or open new one. https://github.com/originell/jpype/

Dhiraj240:
@rtgdk does this issue still need development? I would like to work on it

rtgdk:
@Dhiraj240 I think @techytushar has already pushed #59 for this issue back in Dec 2018. But you can go ahead and check if everything works right now. If not, we would like to migrate to Python3 and Django2 in master. So, we would need a hand to merge the python3 support to master.
Some work is already done in https://github.com/spdx/spdx-online-tools/tree/python3
With this, you can also resolve #58 .

Dhiraj240:
Alright, I will work on this coming weekend and check the files which require changes.

SidharajYadav:
How i can contribute
please guide me
I like to solve this problem

autoantwort:
Can be closed, python 3.7 is currently used.

goneall:
Thanks @autoantwort for reviewing and updating the issues

PR #164, opened 2020-06-07
fix: Replace zip with "itertools.izip".

### LINKED ISSUE #163: Replace itertools.izip
Currently, during the setup process, we get this error:-

`ImportError: cannot import name 'izip' from 'itertools' (unknown location)`

This comes due to izip package becoming obsolete in python 3. In Python 3, we can use zip instead of izip without importing anything to solve this problem.

wryonik:
Since itertools.izip became obsolete in Python 3, this PR replace it by built-in zip() function.

Fixes #163

### DIFF
--- a/src/app/generateXml.py
+++ b/src/app/generateXml.py
@@ -3,7 +3,7 @@ import os
 import re
 import subprocess
 import xml.etree.ElementTree as ET
-from itertools import chain, izip, tee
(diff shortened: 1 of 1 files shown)


rtgdk:
The master is built with python2 and Django=1.11.18, you can add this fix for python3 upgrade #58  #24 , or in the branch `python3`.

PR #167, opened 2020-06-08
Add an option to remove a source

kmehant:
In submitting a new license, we can add more than one sources but according to the previous design, we will be not able to undo this or remove the added source once added. So having an option to remove the added sources would be more UX friendly.

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -224,8 +224,15 @@ $(document).ready(function () {
   });
 </script>
 <script type="text/javascript">
+function removeDiv(elem){
+    $(elem).parent('div').remove();
+} 
+$(document).on('click', '#removeSourceUrl', function(){
+removeDiv(this);
+});
+
 $(document).on('click', '#addSourceUrl', function(){
-    var html = '<div class="col-sm-12 btn-group" style="padding:6px 0; margin-top:2"><input type="text" class="form-control" id="sourceUrl" placeholder="" name = "sourceurl"></div>';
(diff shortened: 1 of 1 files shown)


kmehant:
Fixes #168 

@rtgdk please review :)

src/app/templates/app/submit_new_license.html
```diff
@@ -224,8 +224,15 @@
...
+function removeDiv(elem){
+    $(elem).parent('div').remove();
+} 
+$(document).on('click', '#removeSourceUrl', function(){
+removeDiv(this);
+});
+
 $(document).on('click', '#addSourceUrl', function(){
-    var html = '<div class="col-sm-12 btn-group" style="padding:6px 0; margin-top:2"><input type="text" class="form-control" id="sourceUrl" placeholder="" name = "sourceurl"></div>';
+    var html = '<div class="col-sm-12 btn-group" style="padding:6px 0; margin-top:2"><input type="text" class="form-control" id="sourceUrl" placeholder="" name = "sourceurl"><span class="help-block" id="removeSourceUrl" style="color:red; text-align: left; font-weight: bold; font-style: italic; cursor:pointer;" >- remove</span> </div>';
```
rtgdk, line 235: Can we instead show a "cross" sign (x) next to the row instead of "- remove" in red.

kmehant: @rtgdk I have made the changes.

rtgdk:
@kmehant LGTM. One thing from UX perspective, let's keep the size of cross sign smaller (smaller than the corresponding row) so it looks good and not forcefully placed.
Screenshot 2020-06-10 at 10 20 44 AM

Issue #169, opened 2020-06-08
Update .gitignore


shubhamgupta2956:
Currently, these folders are not in .gitignore
```
src/app/media/
src/spdx-license-matcher/
```
The first folder containers the files uploaded by the user or are generated by any operation by the user (for example- Comparing SPDX RDF Documents).

The second folder is in requirements.txt file.

Apart from that, we can also put `src/src/secret.py` in .gitignore also so that the contributors do not push their credentials accidentally.

PR #170, opened 2020-06-08
Add autogenerated folder in .gitignore.

shubhamgupta2956:
Add `src/app/media/` and `src/spdx-license-matcher/` in `.gitignore` 
Fixes: #169


shubhamgupta2956:
@rtgdk please review.

.gitignore
```diff
@@ -122,3 +122,7 @@ github_utils.py
 
 #Settings secrets
 secret.py
+
+#Autogenerated files
+src/app/media/
+src/spdx-license-matcher/
```
rtgdk, line 128: For me, spdx-license-matcher got downloaded and installed in the virtualenv. Are you getting `src/spdx-license-matcher/` folder? When is this folder created?

rtgdk:
@shubhamgupta2956 Yes, we can add `src/src/secret.py` to this file.

.gitignore
shubhamgupta2956: Yes. Mainly this folder is mentioned in `requirements.txt` file so when we do `pip install -r requirements.txt`, it got installed.

shubhamgupta2956: I guess I forgot to use virtualenv during pip install. Sorry about that. I had removed that file from gitignore as it is installed in `venv` folder which is already present in `.gitingore`

PR #172, opened 2020-06-09
Change source for the spdx-license-matcher to the SPDX github repo

goneall:
This is currently pointing to a fork.  I verified that both are at the same commit level.


Ugtan (commented):
@goneall Since we transferred the repository to spdx, I think we are good to go LGTM.

rtgdk (approved):

Issue #174, opened 2020-06-09
File validation drag/drop doesn't work when filenames have spaces


goneall:
To reproduce, go to the app validate: http://13.57.134.254/app/validate/ and drag and drop any SPDX file with a space in it (e.g. "SPDXTagExample-v2.0 - Copy.spdx").

This results in the error `/var/www/spdxtools/spdx-online-tools/src/app/media/AnonymousUser/1591734127/SPDXTagExample-v2.0%20-%20Copy.spdx does not exist.`

PR #176, opened 2020-06-10
Fix validation for file having whitespaces in their name

shubhamgupta2956:
Currently, when the user uploads a file having white spaces in its name,
it is not validated as the file name is changed due to replacement of
whitespace by "%20". This commit replaces it with whitespaces so that
the correct file can be validated.

Fixes: #174 

Signed-off-by: shubham.gupta2956 <shubham.gupta2956@gmail.com>

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -1153,7 +1153,6 @@ def xml_upload(request):
                             base_url=urljoin(settings.MEDIA_URL, folder+'/')
                             )
                         filename = fs.save(xml_file.name, xml_file)
-                        uploaded_file_url = fs.url(filename)
                         page_id = request.POST['page_id']
                         with open(str(fs.location+'/'+filename), 'r') as f:
                             request.session[page_id] = [f.read(), ""]
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -458,7 +458,7 @@ def validate(request):
                         base_url=urljoin(settings.MEDIA_URL, folder+'/')
                         )
                     filename = fs.save(myfile.name, myfile)
-                    uploaded_file_url = fs.url(filename)
+                    uploaded_file_url = fs.url(filename).replace("%20", " ")
                     """ Call the java function with parameters """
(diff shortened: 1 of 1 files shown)


shubhamgupta2956:
@goneall please review

goneall:
@shubhamgupta2956 Thanks for the fix!

@rtgdk @shubhamgupta2956 I wonder if we have the same issue in other parts of the application where we upload files

shubhamgupta2956:
@goneall I have found this same problem in some more locations also. I had updated the PR for those places also.

goneall:
> I have found this same problem in some more locations also. I had updated the PR for those places also.

@shubhamgupta2956 thanks for catching those - the changes look good to me.

src/app/views.py
```diff
@@ -822,7 +822,7 @@ def convert(request):
                     myfile = request.FILES['file']
                     fs = FileSystemStorage(location=settings.MEDIA_ROOT +"/"+ folder,base_url=urljoin(settings.MEDIA_URL, folder+'/'))
                     filename = fs.save(myfile.name, myfile)
-                    uploaded_file_url = fs.url(filename)
+                    uploaded_file_url = fs.url(filename).replace("%20", " ")
```
rtgdk, line 825: There's one more https://github.com/spdx/spdx-online-tools/blob/f067c36c3822623a257d1d58a5481c9fe71819a6/src/app/views.py#L1156 while uploading xml file

shubhamgupta2956: I had seen this earlier also but did not change it as it does not gives any error while uploading an XML file. It is mainly due to the reason that we are using the `filename` instead of `uploaded_file_url` in the `xml_upload` function. `uploaded_file_url` is of no use in that function. Should I replace it there also or remove that variable?

rtgdk: Sure, go ahead.

rtgdk: You can remove the ununsed variables

shubhamgupta2956:
@rtgdk I have made the required changes

PR #180, opened 2020-06-12
Maintain sensitive data in environment variables

kmehant:
This model of maintaining sensitive data will be beneficial for
1. secure testing in Travis (by using Travis environment variables) or GitHub Workflows
2. container-based setup (passing environment variables to containers would maintain our sensitive data without having them part of the actual code)
3. Easier and cleaner for local setup (as they maintain their own .env file so there will be no need to worry about committing the code with credentials)


kmehant:
Fixes #181

kmehant:
@rtgdk @goneall 
Please review :)

kmehant:
We need to set up those credentials as environment variables in Travis to pass the tests :)

src/src/secret.py
```diff
@@ -1,32 +1,38 @@
...
 def getOauthToolKitAppSecret():
-    return 'OauthAppSecretXXX'
+    return os.environ.get(key="OAUTH_APP_SECRET")
 
 # The method: getAuthCode is important for the license submit tests, given
 # that this authentication code is necessary to generate a github access token.
 # Authentication code generated by client from http://github.com/login/oauth/authorize/
 def getAuthCode():
-    return None
+    return os.environ.get(key="AUTH_CODE")
```
rtgdk, line 38: .env should be added to .gitignore .

rtgdk:
Looks similar to https://github.com/spdx/spdx-online-tools/pull/88, although `python-dotenv` is better than decouple.
@goneall  What do you think about this approach? Is this a priority?

goneall:
I like the approach - it is very similar to #88 

@rtgdk I'll leave it to you which of the two PR's to go with.

I'll try adding the environment variables to Travis and see if I can get the build to pass

goneall:
> I'll try adding the environment variables to Travis and see if I can get the build to pass

No success so far with the environment variables - it may be a problem with escaping.  I'll look into again later this week.

goneall:
I think I found the problem.  From the documentation on setting environment variables "...Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository."

It looks like any encrypted method of providing these values will only work on PR's from this repo.

@rtgdk @kmehant Let me know if you see any way around that.

We could just merge in the PR and see if merged version works.  Unfortunately, the tests will not work for most PR's.

rtgdk:
@goneall I made this PR sometime ago : https://github.com/spdx/spdx-online-tools/pull/156 which was meant to replace travis check with github actions. We can try this method out for setting environment variables and running tests on any PR.
I have never tried setting up the same in travis.yml, but could find onde discussion regarding this : https://github.com/travis-ci/travis-ci/issues/1946
@kmehant Do you have any idea how to proceed?

kmehant:
> I think I found the problem. From the documentation on setting environment variables "...Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository."
> 
> It looks like any encrypted method of providing these values will only work on PR's from this repo.
> 
> @rtgdk @kmehant Let me know if you see any way around that.
> 
> We could just merge in the PR and see if merged version works. Unfortunately, the tests will not work for most PR's.

@goneall Sorry for being a bit late here. Seems like we need to use API Key dealers like [secrethub](https://secrethub.io/docs/guides/travis-ci/) and [algolia](https://secrethub.io/docs/guides/travis-ci/) or set up a [Hasicorp vault](https://hub.docker.com/_/vault) integrated to the Travis pipeline.

kmehant:
> @goneall I made this PR sometime ago : #156 which was meant to replace travis check with github actions. We can try this method out for setting environment variables and running tests on any PR.
> I have never tried setting up the same in travis.yml, but could find onde discussion regarding this : [travis-ci/travis-ci#1946](https://github.com/travis-ci/travis-ci/issues/1946)
> @kmehant Do you have any idea how to proceed?

@rtgdk I guess GitHub Workflows will be great! We can have [encrypted secrets setup](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) too. We can also have more than one workflows now serving the purpose. For Travis, we need to use some secret vaults like [secrethub](https://secrethub.io/docs/guides/travis-ci/) and [algolia](https://secrethub.io/docs/guides/travis-ci/) or set up a [Hashicorp vault](https://hub.docker.com/_/vault) integrated to the Travis pipeline.

goneall:
I just merged #156 - the action is running right now against the master branch.  @kmehant if it is successful, please rebase this PR to pick up the actions and I'll see if I can get the environment variables configured for the Github actions.

kmehant:
> I just merged #156 - the action is running right now against the master branch. @kmehant if it is successful, please rebase this PR to pick up the actions and I'll see if I can get the environment variables configured for the Github actions.

@goneall 
I have rebased and if we can configure [encrypted secrets](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) as the **environment variables** on GitHub Actions we can get the tests passed :)

goneall:
I made some progress - I added secrets for all of the environment variables with the exception of AUTH_CODE. 

Some of the environment variables had to be renamed to work in Github actions.  I added specific review comments for each variable that needs to be renamed.

@kmehant If you could rename the variables.

@rtgdk Can you give me a bit more info for generating the AUTH_CODE.  I take it this is different from the application OAUTH codes.

kmehant:
> I made some progress - I added secrets for all of the environment variables with the exception of AUTH_CODE.
> 
> Some of the environment variables had to be renamed to work in Github actions. I added specific review comments for each variable that needs to be renamed.
> 
> @kmehant If you could rename the variables.
> 
> @rtgdk Can you give me a bit more info for generating the AUTH_CODE. I take it this is different from the application OAUTH codes.

@goneall Review comments are missing on this pull request, can you please add them so that I can change the variable names.

goneall:
@kmehant Looks like I forgot to press the submit button - the comments should be there now

kmehant:
> @kmehant Looks like I forgot to press the submit button - the comments should be there now

@goneall I have the modifications, Please look into them.

goneall:
@kmehant The changes look good.  It doesn't look like the CI build is picking up the secret key for `DJANGO_SECRET_KEY`.  I'll do some experimenting on my private repo to see if I can figure out why.

goneall:
@kmehant I got past the errors in my local clone of the repos - can you push a change to trigger the Github action to rerun?  I don't see a way of manually starting the github actions without pushing a commit.

kmehant:
> @kmehant The changes look good. It doesn't look like the CI build is picking up the secret key for `DJANGO_SECRET_KEY`. I'll do some experimenting on my private repo to see if I can figure out why.



> @kmehant I got past the errors in my local clone of the repos - can you push a change to trigger the Github action to rerun? I don't see a way of manually starting the github actions without pushing a commit.

I triggered Actions.
As far as I remember we can rerun the jobs in the checks section - a button in the top right corner. However I don't see owing to privileges, you might :)

goneall:
@kmehant I now see the rerun task button - somehow I missed that before.

It looks like it is not picking up the `DJANGO_SECRET_KEY` for this repo.

It does pick it up on my local repo, however.   They should be running the exact same code.

I know with Travis, there is an issue where pull requests outside of the account fails to load secret variables for security reasons.  It may be the same case with Actions.

I'll do a bit more research tomorrow and see if I can figure out why it is failing.

goneall:
It looks like secret environment variables will not work for pull requests from forked repositories.  See [this StackOverflow issue](https://stackoverflow.com/questions/58737785/github-actions-empty-env-secrets).

Once we merge the PR, I believe it will work.

@kmehant @rtgdk let me know if you have any ideas on working around this limitation.  I don't think there is an issue with leaving in a default `DJANGO_SECRET`, but the other secrets shouldn't be stored in the clear.

kmehant:
> It looks like secret environment variables will not work for pull requests from forked repositories. See [this StackOverflow issue](https://stackoverflow.com/questions/58737785/github-actions-empty-env-secrets).
> 
> Once we merge the PR, I believe it will work.
> 
> @kmehant @rtgdk let me know if you have any ideas on working around this limitation. I don't think there is an issue with leaving in a default `DJANGO_SECRET`, but the other secrets shouldn't be stored in the clear.

Sorry for being late here,
I never dealt with this type of CI problem before - needing sensitive data for running PR tests. I see there is no such supporting feature is being planned - looking at the GitHub support threads. On the other hand, having credentials available to pull requests could be dangerous. 
@rtgdk Please let me know, how were the credentials provided before. Any thoughts with this problem?

goneall:
@rtgdk Do you know how the tests were passing before?  I looked into the code and couldn't see where the secrets were being set.

rtgdk:
@goneall @kmehant Django secrets key does not have to be correct, it can be any string. So previously, it was picking up `DJANGOSECRETXXX` as the secret and running the test. You can also try this out locally. Other secrets were not being set and the respective tests were not running (you can see `@skipIf` in tests.py and same warning is printed in the logs). Same thing was happening in travis too.

For this PR, the secret key is picked up from `os.environ.get(key="DJANGO_SECRET_KEY")` which is not set in github actions if the PR is raised from forks (this is not supported by travis also I think).

Workaround for this problem (to atleast run basic tests and skip tests requiring github tokens) is to add if-else in `getSecretKey` function - if the os.environs returns empty string, we should return a non empty string (maybe "NON_EMPTY_DJANGO_SECRET")

src/src/secret.py
```diff
@@ -1,32 +1,38 @@
...
-    return 'GHKEYXXX'
+    return os.environ.get(key="ONLINE_TOOL_GITHUB_KEY")
 
 def getGithubSecret():
-    return 'GHSECRETXXX'
+    return os.environ.get(key="ONLINE_TOOL_GITHUB_SECRET")
 
 def getSecretKey():
-    return 'DJANGOSECRETXXX'
+    return os.environ.get(key="DJANGO_SECRET_KEY")
```
rtgdk, line 14: Need to add if-else here to run basic tests.

kmehant:
@rtgdk @goneall 
Please review!

goneall:
@rtgdk Thanks for investigating the issues and thanks @kmehant for the fix!

Once we merge this, the merge should use the stored secrets and run the additional tests.

@rtgdk if it looks good to you, please go ahead and merge

src/src/secret.py
```diff
@@ -1,32 +1,50 @@
...
 
 def getGithubSecret():
-    return 'GHSECRETXXX'
+    return os.environ.get(key="ONLINE_TOOL_GITHUB_SECRET")
+
 
 def getSecretKey():
-    return 'DJANGOSECRETXXX'
+    return "NON_EMPTY_DJANGO_SECRET" if os.environ.get(key="DJANGO_SECRET_KEY") == "" \
+        else os.environ.get(key="DJANGO_SECRET_KEY", failobj="NON_EMPTY_DJANGO_SECRET")
```
rtgdk, line 18: I found `os.getenv(key,default)` to do the same thing this big if-else is doing. Should we change it to that?

rtgdk:
@kmehant LGTM, left one comment. Will merge after that change is pushed..

src/src/secret.py
kmehant: @rtgdk actually the environment variable is holding an empty string `""` which is provided by GitHub to outside PRs, on the other hand, I observed os module utilizes default value only when the environment variable is `None` and not for `""` (empty string). 
I feel this is the only fix, thoughts?

rtgdk: @kmehant Makes sense. Merging.

rtgdk:
@goneall @kmehant Looks like the github tests ran fine on merge, there were some failed tests but I think that is related to https://github.com/spdx/spdx-online-tools/pull/154.

PR #182, opened 2020-06-13
Increase modal view to remove overflowing errors.

shubhamgupta2956:
I have tested for many cases and this size of modal seems to be most appropriate for any type of errors.

Current modal view:-
2020-06-13_17-09-24-1920x1080

New modal view:-
2020-06-13_17-08-44-1920x1080

PS- Since the CSS properties of this modal are coming from bootstrap, I had to use inline CSS for it. 

Signed-off-by: shubhamgupta2956 <shubham.gupta2956@gmail.com>

### DIFF
--- a/src/app/templates/app/modal.html
+++ b/src/app/templates/app/modal.html
@@ -1,6 +1,6 @@
 <!-- Modal -->
 <div id="myModal" class="modal fade" role="dialog">
-  <div class="modal-dialog">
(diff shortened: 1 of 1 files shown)


shubhamgupta2956:
@rtgdk please review.

rtgdk:
@shubhamgupta2956 LGTM.

PR #185, opened 2020-06-14
Add reset button to submit new license and namespace section

Ugtan:
Added a reset button to reset the form in the submit license new request and submit namespace section.

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -159,7 +159,7 @@
       </div>
     </div>
     <hr>
-		<input type="hidden" id="cfileformat" name="submit" value="">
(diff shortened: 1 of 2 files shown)


rtgdk:
@Ugtan Changes look fine.
@goneall Do we need this reset functionality in the form? From UX perspective, reset buttons can be dangerous and if pressed by mistake can't be undone -- all your work is gone!. Even `Cmd+Z(Ctrl+Z)` does not work after reset, it works in usual interactions with the form.
Having two buttons at the bottom of a form clutters up the interface and makes it harder for users to clearly see their next step.

goneall:
I haven't heard any feedback from the users that we need a reset, so perhaps we should leave this out for the reasons @rtgdk described above.

Ugtan:
Okay. No issues. :)

I'm closing the PR.

PR #188, opened 2020-06-22
Write production and developement container configurations 

### LINKED ISSUE #183: Dockerize spdx-online-tools
Dockerizing will enable a smoother flow of execution and development.

kmehant:
## Covers

1. Containerize SPDX for development purpose: no hassle to install dependencies now, simply run `docker-compose -f docker-compose.dev.yml up --build`
2. Containerize SPDX for production purpose: with WSGI (for concurrency) and supervisor (extra layer for more uptime and concurrency). simply run `docker-compose -f docker-compose.prod.yml up --build`
3. docker-compose configs: no hassle of remembering the container spin-up configurations

Fixes #183


kmehant:
@rtgdk @goneall Please review!

rtgdk:
@kmehant So quick! 😄 
This is an important PR, give me some time to review and check.
Can you provide the instruction to setup and deploy this? Also, the thinking that went behind to do it like this. I thought this would be a summer project but the amount of changes looks small.
@goneall would be the one to deploy the tools online.

kmehant:
> @kmehant So quick! 😄
> This is an important PR, give me some time to review and check.
> Can you provide the instruction to set up and deploy this? Also, the thinking that went behind to do it like this. I thought this would be a summer project but the amount of changes looks small.
> @goneall would be the one to deploy the tools online.

@rtgdk efforts were part of the making of my [proposal](https://drive.google.com/file/d/1h9E_LWDohZO3uflAgEFnfCqjgXrzO91Z/view?usp=sharing) and my previous experience so it was quick. I guess, there is much more to be done - you can find my thought process and goals from my proposal [here](https://drive.google.com/file/d/1h9E_LWDohZO3uflAgEFnfCqjgXrzO91Z/view?usp=sharing), but this PR forms the base for container images and local setup. 

#### Instructions
You need to have [docker desktop](https://docs.docker.com/desktop/) installed on your machine for the container environment. 
As this pull request also covers docker-compose configs so it is as simple as running (for development build)
```
docker-compose -f docker-compose.dev.yml up --build
```
(for production build)
```
docker-compose -f docker-compose.prod.yml up --build
```
@goneall mentioned that we will be using a PostgreSQL for online tools but haven't found any setup in the codebase. I could only see Redis and sqlite3 used.

Also there is another change required before running the docker-compose
```
r = redis.StrictRedis(host='localhost', port=6379, db=0)
```
Its good if we can get Redis host form environment variables so it will be helpful for the container setup. For now please change the host to `redis` from `localhost` as our container utilizes the docker-compose network to communicate with the Redis container.

dockerfile.dev
```diff
@@ -0,0 +1,26 @@
+# Use the official python 2 base image
+FROM python:2
```
Ugtan, line 2: How about using an alpine based docker image for python here?

dockerfile.prod
```diff
@@ -0,0 +1,54 @@
...
+FROM base as linter 
+
+# Setting current work directory to /spdx
+WORKDIR /spdx
+
+# Copy all files into the container
+COPY . .
+
+# Installing pylint and linting the python code
+RUN pip install pylint && pylint -E *.py **/*.py || printf "\n SPDX online tools has got some errors! Continuing the build for now\n\n"
```
Ugtan, line 25: What if we add pylint or any other linter in the GitHub actions and make sure each pull request that we merge is error-free? That way I don't think we would have to check it here.

What do you think?

dockerfile.dev
```diff
@@ -0,0 +1,26 @@
...
+
```
Ugtan, line 3: Set up a environment variable to avoid writing pyc files?

dockerfile.prod
```diff
@@ -0,0 +1,54 @@
...
+
+# Run the tests
+RUN python src/manage.py test && printf "Success!"
+
+# -----------------------------
+
+# Container for for SPDX Online tools
+FROM base as production
+
+# Setting current work directory to /bassa
```
Ugtan, line 35: `/spdx` here instead of `/bassa`

supervisor_api.conf
```diff
@@ -0,0 +1,6 @@
+[program:spdx]
+directory=/spdx/src
+command=gunicorn -c /spdx/wsgi_conf.py src.wsgi:application --bind 0.0.0.0:8000
+autostart=true
+autorestart=true
+startretries=5
```
Ugtan, line 6: How about adding a stdout logfile here?

Ugtan:
Hello @kmehant, I have added some comments. Could you please take a look at it. Also, have you tried it locally in your machine to see if all the functionalities are working fine?

Also, I found [this](http://docs.gunicorn.org/en/stable/faq.html#how-do-i-avoid-gunicorn-excessively-blocking-in-os-fchmod) in the gunicorn docs which might affect the performance in production.

> we will be using a PostgreSQL for online tools but haven't found any setup in the codebase. I could only see Redis and sqlite3 used.

I suggested it in my proposal to use Postgres to store data on a database that will be running on another server instance. This will help the app to scale horizontally and use a full-fledged database. We will probably have to implement it from scratch.

Just to let you know I have also submitted a proposal for the same project and I had a word with @goneall and I expressed my interest in contributing to the project as well. He said we will assign some tasks between the three of us. I think he is a little busy with work.

dockerfile.dev
kmehant: @Ugtan I went through this article and the reasoning behind not using `alpine` [link](https://pythonspeed.com/articles/alpine-docker-python/). thoughts?

dockerfile.dev
kmehant: @Ugtan Any reason why to avoid writing pyc files in a container?

dockerfile.dev
Ugtan: @kmehant I guess it is better not to use alpine for the reason mentioned in the article. Thanks for the article tho.

dockerfile.dev
Ugtan: To reduce the image size. Maybe you can try adding it to see it makes any difference in the overall size of the image?

kmehant: I have added it to docker-compose so users can have it on/off on their preference. 
Also added `.dockerignore` so cache, test files, git files, and other such do not creep into the image.

Ugtan: Great!

goneall:
Thanks @kmehant for the PR and thanks @Ugtan for the review and comments!  It's great to see the contributions and collaboration :)

In terms of the production changes, there is a very old production branch which contains 2 commits for the changes in the production code when it was first set up: https://github.com/spdx/spdx-online-tools/tree/productionmods

I just looked and it doesn't look like it contains the configuration changes for PostgreSQL, but it does include the NGINX configuration used.

I'll be spending some more time helping out on the production in the next couple of weeks.

Once I get the environment variables working for the github actions (PR #180 ), I'll take a look at the production environment and update the production branch with the current configurations.

goneall:
Just FYI - I check the production instance and the only changes are the following:

- Installation of psycopg2
- Setting debug to false
The following change to the settings.py:

```
DATABASES = {
     'default': {
       'ENGINE': 'django.db.backends.postgresql_psycopg2',
       'NAME': 'spdx_db',
       'USER': 'spdx',
       'PASSWORD':'********',
       'HOST':'localhost',
       'PORT':'5432'
   }
```

The database just runs locally.

goneall:
I'm just getting started experimenting with the docker file.

I'm getting the following error:

```
ERROR: The Compose file '.\docker-compose.dev.yml' is invalid because:
services.spdx.environment contains {"PYTHONDONTWRITEBYTECODE": 1}, which is an invalid type, it should be a string
```
when running the command `docker-compose -f docker-compose.dev.yml up --build` on a Windows 10 computer

I tried adding quotes around the `'1'` but got the same error.

Removing the environment section resolved the issue.

@kmehant @Ugtan Let me know if you have any suggestions.

goneall:
I was able to get the docker image up and running :)

Testing out a few of the features, it looks like we need to add a Java JVM to the configuration.

Running the verify command, I got the following exception:

```
OSError: [Errno 2] No such file or directory: '/usr/lib/jvm'
```

kmehant:
> I was able to get the docker image up and running :)
> 
> Testing out a few of the features, it looks like we need to add a Java JVM to the configuration.
> 
> Running the verify command, I got the following exception:
> 
> ```
> OSError: [Errno 2] No such file or directory: '/usr/lib/jvm'
> ```

@goneall 
* ✅ Fixed PYTHONDONTWRITEBYTECODE
* ✅ Added JDK8 to both development and production containers. Please test it out.

goneall:
Thanks @kmehant !

That fixed the two issues I ran into.

There is a new issue I ran into while trying to use the Verify function (which uses the Java code through JPype).  I'm getting the following error:

```
Traceback (most recent call last): File "/spdxonlinetools/src/app/views.py", line 463, in validate retval = verifyclass.verify(str(settings.APP_DIR+uploaded_file_url)) File "/usr/local/lib/python2.7/site-packages/jpype/_jpackage.py", line 62, in __call__ raise TypeError("Package {0} is not Callable".format(self.__name)) TypeError: Package org.spdx.tools.Verify.verify is not Callable
```

When I get some time, I can do some debugging.  I suspect it may be a permission problem.  I did verify the Java code (src/tool.jar) was working properly standalone.  When I get some time, I'll try the code in the branch outside of the Docker image to make sure it isn't an unrelated issue.

goneall:
I spent some time trying to debug the JPype TypeError problem described above without success (so far).

Here's what I tried/verified:

- If I run the Python server code locally outside the container, it works without error
- Running the Java tools JAR inside the container using the container command line works
- I updated the JVM version to match the version I was using on my local machine (JVM 13), did not resolve the issue
- I verified the Java tools JAR file had execution permissions
- I verified the requirements.txt file specified the version of JPype1 - so my local machine and the remove machine should have the same version of the library
- I found a [stack overflow article](https://stackoverflow.com/questions/21006124/using-custom-java-classes-with-jpype-package-not-callable-error) which seemed to have an unrelated issue, but the solution did not work - JClass was not recognized

There are a few more google results I can try - some suggesting checking the classpath (e.g. https://github.com/jpype-project/jpype/issues/173)

@rtgdk Let me know if you have any ideas - I know you've been successful solving JPype errors.  When I get some time, I'll do some more research/attempts.

Ugtan:
Hello @goneall, I have encountered this error before. It occurs when the classpath is not correct. Simply changing the `JAR_ABSOLUTE_PATH` in `settings.py` to `JAR_ABSOLUTE_PATH =  os.path.join(os.path.abspath("."), "src", "tool.jar")
` fixes the issue. Also, everything seems to be working fine except the check license feature which uses redis. As there is some issue with the redis host.

kmehant:
> Hello @goneall, I have encountered this error before. It occurs when the classpath is not correct. Simply changing the `JAR_ABSOLUTE_PATH` in `settings.py` to `JAR_ABSOLUTE_PATH = os.path.join(os.path.abspath("."), "src", "tool.jar") ` fixes the issue. Also, everything seems to be working fine except the check license feature which uses redis. As there is some issue with the redis host.

Have tried changing the hostname for Redis?

rtgdk:
@goneall @Ugtan Yeah, changing the Jar path would fix the issue. Although, I would suggest to make it
`JAR_ABSOLUTE_PATH =  os.path.join(BASE_DIR,"tool.jar")`
This will be a common fix and would work outside docker too.
@kmehant Can you include this change in your PR?

goneall:
I got everything working locally.  I had to rename Redis in 3 locations:

- spdx_license_matcher/build_licenses.py (in the spdx-license-matcher repo)
- spdx_license_matcher/matcher.py (in the spdx-license-matcher repo)
- src/app/util.py

I'll go ahead and merge this PR and start organizing a list of production oriented tasks as separate issues

Issue #216, opened 2020-08-15
Need to configure License Text Diff


goneall:
This is an issue for the production instance.

Issues submitted to update the license text point to https://github.com///blob/master/3y7usicope.png which seem to be missing github account.

See https://github.com/spdx/TEST-LicenseList-XML/issues/77 for an example.

This likely needs to be configured.

goneall:
@Ugtan Do you know what needs to be configured for the license diff repo?

Gary

Ugtan:
@goneall Please check out this comment for info https://github.com/spdx/spdx-online-tools/pull/136#issuecomment-533274704. Also, Instead of hardcoding these, we should take these inputs from the environment variables just like we are using the `spdx-prod.env` in production.

goneall:
Resolved with PR #217

PR #217, opened 2020-08-18
Add configuration for screenshot repository

goneall:
This is to resolve issue #216 

### DIFF
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -19,7 +19,6 @@ from django.contrib.admin import widgets
 
 from app.models import UserID, LicenseNamespace, OrganisationName
 from app.widgets import RelatedFieldWidgetCanAdd
-from django.conf import settings
 
 OSI_CHOICES = (
     (0, "-"),
@@ -84,8 +83,6 @@ class LicenseRequestForm(forms.Form):
     comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
-    repositoryNameWithOwner = settings.DIFF_REPO_WITH_OWNER
-    diffRepoGithubToken = settings.DIFF_REPO_GIT_TOKEN
 
 
 class LicenseNamespaceRequestForm(forms.ModelForm):
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -73,8 +73,6 @@ class LicenseRequestForm(forms.Form):
             self.email = kwargs.pop('email')
         else:
             self.email = ""
-        if 'githubtoken' in kwargs:
-            self.githubtoken = kwargs.pop('githubtoken')
         super(LicenseRequestForm, self).__init__(*args,**kwargs)
         self.fields["userEmail"] = forms.EmailField(label='Email', initial=self.email)
 
@@ -86,7 +84,8 @@ class LicenseRequestForm(forms.Form):
     comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
-    repositoryNameWithOwner = settings.DIFF_REP_WITH_OWNER
+    repositoryNameWithOwner = settings.DIFF_REPO_WITH_OWNER
+    diffRepoGithubToken = settings.DIFF_REPO_GIT_TOKEN
 
 
 class LicenseNamespaceRequestForm(forms.ModelForm):
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -19,6 +19,7 @@ from django.contrib.admin import widgets
 
 from app.models import UserID, LicenseNamespace, OrganisationName
 from app.widgets import RelatedFieldWidgetCanAdd
+from django.conf import settings
 
 OSI_CHOICES = (
     (0, "-"),
@@ -72,6 +73,8 @@ class LicenseRequestForm(forms.Form):
             self.email = kwargs.pop('email')
         else:
             self.email = ""
+        if 'githubtoken' in kwargs:
+            self.githubtoken = kwargs.pop('githubtoken')
         super(LicenseRequestForm, self).__init__(*args,**kwargs)
         self.fields["userEmail"] = forms.EmailField(label='Email', initial=self.email)
 
@@ -83,6 +86,7 @@ class LicenseRequestForm(forms.Form):
     comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
+    repositoryNameWithOwner = settings.DIFF_REP_WITH_OWNER
 
 
 class LicenseNamespaceRequestForm(forms.ModelForm):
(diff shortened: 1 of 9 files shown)


goneall:
This PR is almost, but not quite working.  @Ugtan @rtgdk I could use some help from someone more knowledgeable in Python/Django than myself.

I was able to configure the parameters and pass them to JavaScript, but I am now getting a 404 error when posting to the Github API from https://github.com/spdx/spdx-online-tools/blob/b3669e2cf0762796a8e4a065a94a3d8873fe8332/src/app/static/js/editor/postscreenshot.js#L44

The URL it is posting to is `https://api.github.com/repos/spdx/licenseRequestImages/contents/t82ynv8qrv.png`

goneall:
I'm also passing the Github token from Python to JavaScript - @rtgdk @Ugtan  wouldn't this login information already be available in JavaScript?  If so, what variable is it stored in?

Ugtan:
@goneall First of all, the github token in the postscreenshot.js file is the token of the owner of the repository and not the user logging into the app(sorry for a vague name) i.e spdx in our case. I think we should use a dummy github account to store images instead of using a spdx repository for security purposes. We could also switch to third party applications like Imgur for storing diff images if we decide not to store it on github(But, that would also require a secret key). 

Secondly, We could store the token in an environment variable in python, defined it in the settings.py file, and pass it to the template using context processor or else use dotenv with node to get environment variables.

Let me know if you need any help. 

Thanks

goneall:
@Ugtan Can you provide a bit more context on how to obtain the token?  In GitHub, I can get an SSH key and create application access keys, but I'm not sure of the specific process needed to obtain the token for this application.

rtgdk:
@goneall @Ugtan I finally understood the working of github api! The docs and examples are a mess :(
I tried github api - https://developer.github.com/v3/repos/contents/#create-or-update-file-contents as being done in postscreenshot.js, used a personal access token and commited files in https://github.com/rtgdk/test_github_api using the token.
@goneall From what I understood, Personal access token are not available for organisation, so you need to create a repo on your personal account, then create a personal access token with access for write to repository(here's the guide for that - https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token). Save this personal token in env variable - `spdx-prod.env` in production and pass this token in `postToGithub` function instead of user's github token -`githubToken`. Also, use your own account's name in `repositoryNameWithOwner`. One more thing I noticed, we need to base64 encode the file created and send it as "content" params. 
Here's the curl request I used while testing

> curl   -X PUT   -H "Accept: application/vnd.github.v3+json" -H  "Authorization: Token $P-AUTH-TOKEN" "https://api.github.com/repos/rtgdk/test_github_api/contents/test.txt"   -d '{"message":"message","content":"Y29udGV4dA=="}'

`Y29udGV4dA==` is base64 encoding of work "content"

So pseudo code for uploading the file from postscreenshot.js would look like

>params = Json.stringify({"content" : base_64encode(filename), "message" : "Uploading - " + filename})
apiurl =
    "https://api.github.com/repos/" +
    repositoryNameWithOwner +
    "/contents/" +
    filename;
  xhr.open("PUT", apiurl);
  xhr.setRequestHeader("Authorization", "Token " + personalAccessToken);
  xhr.send(params);


1. https://docs.github.com/en/rest/reference/repos#create-or-update-file-contents (better api examples)
2. https://github.community/t/how-to-upload-a-file-using-github-api/14031/3 (see the comment by `greenandgreen`)
3 [Future] Also, if by any chance the filename already exists in the repo, then we need to follow some different steps and params  (We need to send "sha" param as well)

Let me know if you need any help with this.
@Ugtan Please correct me if I am wrong anywhere.

src/app/templates/app/submit_new_license.html
```diff
@@ -552,6 +552,8 @@
 
 /* generate diff of input license text and spdx license list license text of closely matched licenses */
 async function generate_text_diff(base, newtxt){
+    var repositoryNameWithOwner = "{{ form.repositoryNameWithOwner }}";
+	var diffRepoGithubToken = "{{ form.diffRepoGithubToken }}";
```
rtgdk, line 556: nit: indent, although it doesn't matter in html

src/app/forms.py
```diff
@@ -83,6 +84,8 @@ def __init__(self, *args, **kwargs):
     comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
+    repositoryNameWithOwner = settings.DIFF_REPO_WITH_OWNER
+    diffRepoGithubToken = settings.DIFF_REPO_GIT_TOKEN
```
rtgdk, line 88: Not sure whether passing the secret key in the forms is a good idea. We should have the screenshot function on the backend side for more security, but let's go with this now.

rtgdk (approved):
LGTM overall, please check that you can't see the secret token in form metadata on html.

goneall:
Added issue to track the performance improvement: https://github.com/spdx/LicenseListPublisher/issues/73

Ugtan (commented):
@goneall Passing a github token through forms is not at all a good solution IMO.

Agreed with @rtgdk to take a screenshot from the backend. It will be a way easier task than passing sensitive data from python to js. +1 to the base64 encoding of the image before sending it to the GitHub api and taking into consideration for same name file as GitHub will automatically overwrite the file if it is already present in the repository.

goneall:
@rtgdk good point on the the secret key in JavaScript.  It looks like the secret would be exposed in the JavaScript and HTML - I'll look into moving this to the backend.

goneall:
@Ugtan @rtgdk What about just uploading the image to the backend media folder and creating a link to that in the issue?  I'm not very proficient in Django, but it seems like that would be supported and easier than storing in a separate repository.

rtgdk:
@goneall Yeah, we can store the files in the media folder from javascript but the media folder is ephemeral (not sure if the current aws instance always remains up). I think docker is ephemeral - meaning that any changes to the filesystem while the container is running only last until that container is shut down or restarted. Saw this kind of problem while hosting on heroku (More info - https://help.heroku.com/K1PPS2WM/why-are-my-file-uploads-missing-deleted). The solution is to upload the files in s3 or some other permanent photo upload system like imgur.
Also, one more problem is we currently don't  serve the media folder in non debug mode - https://github.com/spdx/spdx-online-tools/blob/master/src/src/urls.py#L60, because all the files uploaded by users are stored there.
Another solution I see is to use Github.js (https://www.audero.it/blog/2015/10/17/upload-files-on-github-using-github-js/#uploading-files-with-github-js), haven't tried it out yet.

I'll research more on this topic and let you know.

rtgdk:
@goneall  @Ugtan I have implemented backend upload of diff images, turns out the images were already base64encoded  (See `canvas.toDataURL();` in `postscreenshot.js`). I created a simple backend api which is called through ajax request by `postToGithub`. Please review.


Another approach that @goneall suggested, was to store the files in the database itself as blobs. Django supports this feature but in it's [official documentation](https://docs.djangoproject.com/en/3.1/ref/models/fields/#binaryfield) and many other articles, they have said that storing static files in the database is a bad idea (there are performance issues with having large data in a column mysql/postgres). I have opened an issue #218  to research more on this and track this request. If we find that storing the files in the database is the way to go, I'll open a PR for that.

goneall:
@rtgdk I just tested out the new code and the image uploads fine, but the issue submitted no longer contains the link to the image.

Before the change, the issue includes the link - see https://github.com/spdx/TEST-LicenseList-XML/issues/79

After the change, the issue no longer include the image link - see https://github.com/spdx/TEST-LicenseList-XML/issues/80

src/app/static/js/editor/postscreenshot.js
```diff
@@ -30,19 +27,56 @@ async function getFileData(imgUrl) {
...
+        type: "POST",
+        enctype: 'multipart/form-data',
+        url: "/app/post_to_github/",
+        dataType: 'json',
+        timeout: 600000,
+        data: data,
+        success: function (data) {
+            var githubCode = data.statusCode;
+            if(githubCode == '201'){
+              fileurl = data.fileurl
```
goneall, line 41: If this is an asynchronous call, the function will return null for fileurl if it returns before the ajax call is returned.

A simple fix would be to add  `async: false`

goneall:
@rtgdk @Ugtan I just added a fix for the missing URL link.  I'll go ahead and merge these changes and start working on a couple other minor UI improvements.

PR #219, opened 2020-09-04
Configure SSL for the production docker images

goneall:

Sets up SSL using certbot and lets-encrypt


goneall:
This has not yet been tested.  The changes are based on the article [Nginx and Let’s Encrypt with Docker in Less Than 5 Minutes](https://medium.com/@pentacent/nginx-and-lets-encrypt-with-docker-in-less-than-5-minutes-b4b8a60d3a71).

It is currently configured to encrypt and proxy to the old implementation.  Once we have tested the new production instance, I'll reconfigure to not use staging for the certificates and proxy to the new docker production image instead of the old service.

goneall:
This is now tested and live at http://tools.spdx.org and https://tools.spdx.org

goneall:
@swinslow @kmehant This now seems to be working.

Let me know if you see any possible improvements.

rtgdk (approved):

Issue #222, opened 2020-09-07
spdx-license-matcher not current in production container


goneall:
Rebuilt the production container by running `pip-install -r requirements.txt` then running `docker-compose -f docker-compose.prod.yml build`.  The image produced has the spdx-license-matcher as a subdirectory of spdx/src.  Examining the sub-directory, the image is behind master by a commit (2 file changes).  Expected behavior is for the package to be current as of the time of the build.

PR #229, opened 2020-09-07
Resolve issue #222 - stale spdx-license-matcher

goneall:


goneall:
This fix requires the spdx-license-matcher be tagged and the requirements file to be updated to reference the specific spdx-license-matcher tag.  This forces docker to avoid the cached layer with the requirements when the spdx-license-matcher is updated.

Issue #227, opened 2020-09-07
Compare and Spreadsheet result file types incorrect


goneall:
The file extensions for the spreadsheets should be "xls" instead of "xlsx".

goneall:
Resolved by PR #231

Issue #226, opened 2020-09-07
Compare for the SPDX 2.2 examples fail


goneall:
When comparing the [tag/value](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXTagExample-v2.2.spdx) and [RDF](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXRdfExample-v2.2.spdx.rdf.xml) SPDX 2.2 example files, the compare fails with the following error:

![image](https://user-images.githubusercontent.com/378172/92413221-f1ac0580-f103-11ea-865c-ffeca40cecaa.png)

When running the command line using the Jar, the compare succeeds.

goneall:
Resolved by PR #231

Issue #215, opened 2020-08-15
Change message when submitting license update request


goneall:
When there is a near match of a license when submitting a new license request, the user is prompted to compare the licenses. 
 When the user chooses the option that the licenses should match, an issue is submitted requesting the license to be updated.

The message displayed seems to indicate a new license request is submitted.  The message be changed to something like "A request has been sent to update the license text to match".

Ugtan:
> When there is a near match of a license when submitting a new license request, the user is prompted to compare the licenses.
When the user chooses the option that the licenses should match, an issue is submitted requesting the license to be updated.

@goneall If the license is a close match, the user is prompted to choose if the changes made by the user is substantial or if they think the licenses should match. 

If the changes are substantial a new request i.e GitHub issue is created so "A new license request" makes more sense, thoughts?

If the license should match(the case you are talking about), a new pull request is added to the desired repository. If you want we could update the pull request title or body to display the message "A request has been sent to update the license text to match".

goneall:
> If the changes are substantial a new request i.e GitHub issue is created so "A new license request" makes more sense, thoughts?

Their response indicates that the changes are not substantial and the licenses should match, so the issue created is correct.  It is just the message displayed back to the user isn't correct.  The message states that a new license request is submitted when in fact it is an issue requesting the licenses to match.

Only the message displayed to the user needs to be updated.  I thought the issue submitted to github was fine.

goneall:
Resolved with PR #231

PR #231, opened 2020-09-07
Fix various issues found in production

goneall:
Resolves issue #211 
Resolves issue #215 
Resolves issue #226 
Resolves issue #227 

Note that there are separate commits for each of the issues if you prefer to review individually.

Also note that the development docker-compose file now requires a .env file at the root of the project to hold sensitive environment variables.


Ugtan (approved):
LGTM, changes looks good to me @goneall

Issue #237, opened 2020-09-18
How do we "authorize" users to archive license requests?


goneall:
The archive license requests is only allowed for authorized users.

Since the logins are now done by Github, creating a superuser for this function won't work.

How do we enable an existing user to be authorized for these functions?

goneall:
I found that code that checks the permissions at: 

https://github.com/spdx/spdx-online-tools/blob/415fcecd01a9fa08036dbe53ad392c476de59c4c/src/app/utils.py#L56

When I use the production instance at http://13.57.134.254/app/ my username is goneall9ea2cb6faa12445f and I am authorized.

On the test instance at http://52.32.53.255/app, my username is goneall and I am not authorized.

rtgdk:
@goneall It checks if the user has collaboration access to https://github.com/spdx/tools-python/. Can you check your github login username from superadmin on the test instance (Not sure if `goneall9ea2cb6faa12445f` is github login username or django db username)? I am not a superadmin so I couldn't check. (Also, maybe we need to change the repo to https://github.com/spdx/license-list-XML?)

Also, I think we need to change who can archive license requests feature - the person who submitted the requests, and spdx team members. I think we can assign a role to spdx team members and that should be checked instead of collabooration access. Django provides RBAC - but some super admin has to assign the roles (manual process).

goneall:
After reading through the code, it looks like the fix for authorizing the archive function was inside the post code.  I submitted PR #237 to resolve.

PR #239, opened 2020-09-21
Fix archive authorization - resolves issue #237

goneall:

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -1291,30 +1291,32 @@ def archiveRequests(request, license_id=None):
     returns archive_requests.html template
     """
     context_dict = {}
+    if request.user.is_authenticated():       
+        user = request.user
+        github_login = user.social_auth.get(provider='github')
+        if utils.checkPermission(user):    
+            archive = request.POST.get('archive', True)
+            license_id = request.POST.get('license_id', False)
+            context_dict['authorized'] = "True"
+    else:
+        github_login = None
     if request.method == "POST" and request.is_ajax():
-        if request.user.is_authenticated():       
-            user = request.user
-            if utils.checkPermission(user):    
-                archive = request.POST.get('archive', True)
-                license_id = request.POST.get('license_id', False)
-                context_dict['authorized'] = "True"
-                if license_id:
-                    LicenseRequest.objects.filter(pk=license_id).update(archive=archive)
-        else:
+        if not request.user.is_authenticated():
             if (request.is_ajax()):
                 ajaxdict = {}
                 ajaxdict["type"] = "auth_error"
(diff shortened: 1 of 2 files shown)


goneall:
@anshuldutt21 - pls. review

rtgdk:
@goneall I am not sure this will fix anything. On further research, on production instance - http://13.57.134.254/app/ anybody can archive any license even without login(archive button is shown to everybody). (since the code deployed is very old) Whereas on the test instance, http://52.32.53.255/app  only for authorised people(context_dixt["authorised"] and github login is true), we show the archive/unarchive button. This was done in https://github.com/spdx/spdx-online-tools/pull/154 (which I don't think is deployed on production instance)

src/app/views.py
```diff
@@ -1291,30 +1291,32 @@ def archiveRequests(request, license_id=None):
     returns archive_requests.html template
     """
     context_dict = {}
+    if request.user.is_authenticated():       
+        user = request.user
+        github_login = user.social_auth.get(provider='github')
+        if utils.checkPermission(user):    
+            archive = request.POST.get('archive', True)
+            license_id = request.POST.get('license_id', False)
```
rtgdk, line 1299: This will error out for `GET` request since request.POST will not exist, not sure though

rtgdk: Also, we don't need these lines, not used anywhere
```
archive = request.POST.get('archive', True)
license_id = request.POST.get('license_id', False)
```

goneall:
@rtgdk I'm not completely sure how this function works - I think it may be called in 2 scenerios:

- When retrieving the list of licenses using `get`
- When setting the archive flag in the database when the archive button is pushed using `post`

I'm not sure about the `ajax` flags.

I did run some tests on my local system and the code changes did fix the problem.

Without these code changes, `authorized` never gets set and `utils.checkPermission(user)` never gets called.

rtgdk:
@goneall yeah, you are right. This function is called for both get and post request. And you right to set the `authorized` flag. But now anyone who can login to the spdx online tool - can do a post request(ajax call) to archive/unarchive the license (He won't see the archive/unarchive button on the UI but will be able to achieve it through the post request)

rtgdk:
@goneall I have the required changes in [this commit](https://github.com/spdx/spdx-online-tools/pull/239/commits/0b72f89b24f1ac8529d9830564f8b3c5ed73cbb7). Also, regarding the github collaborator check, it currently checks for https://api.github.com/repos/spdx/tools-python/collaborators/, should we change it to spdx/license-list-XML?

goneall:
Thanks @rtgdk - I make one more commit to change the check to work against the license-list-XML repo rather than the python tools repo.

If it looks good to you, go ahead and merge.

PR #245, opened 2020-10-10
Use the new tools-java adding support for JSON, YAML, XLSX, and XML file formats

goneall:
The new Java tools library has been significantly re-written and adds support for additional JSON file.

Note: There is some loss functionality in the convert function.  It no longer supports converting to HTML.

The HTML and JavaScript UI has been update to reflect the new files supported.

Also note that the default spreadsheet format is now XLSX rather than XLS (although XLS is also supported).

### DIFF
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -37,7 +37,7 @@ import datetime
 import xml.etree.cElementTree as ET
 
 from traceback import format_exc
-from os.path import abspath, join
+from os.path import abspath, join, sep, splitext
 from time import time
 from requests import post
 from json import dumps, loads
@@ -99,8 +99,11 @@ def validate(request):
             uploaded_file_path = str(query.file.path)
             try :
                 if request.FILES["file"]:
+                    formatstr = request.POST["format"]
+                    serFileTypeEnum = jpype.JClass("org.spdx.tools.SpdxToolsHelper$SerFileType")
+                    fileformat = serFileTypeEnum.valueOf(formatstr)
                     """ Call the java function with parameter"""
-                    retval = verifyclass.verify(uploaded_file_path)
+                    retval = verifyclass.verify(uploaded_file_path, fileformat)
                     if (len(retval) > 0):
                         result = "The following error(s)/warning(s) were raised: " + str(retval)
                         returnstatus = status.HTTP_400_BAD_REQUEST
@@ -147,14 +150,38 @@ def extensionGiven(filename):
         return False
 
 def getFileFormat(to_format):
-    if (to_format=="Tag"):
+    if (to_format=="TAG"):
         return ".spdx"
-    elif (to_format=="RDF"):
-        return ".rdf"
-    elif (to_format=="Spreadsheet"):
+    elif (to_format=="RDFXML"):
+        return ".rdf.xml"
+    elif (to_format=="XLS"):
         return ".xls"
-    elif (to_format=="HTML"):
-        return ".html"
+    elif (to_format=="XLSX"):
+        return ".xlsx"
+    elif (to_format=="JSON"):
+        return ".json"
+    elif (to_format=="YAML"):
+        return ".yaml"
+    elif (to_format=="XML"):
+        return ".xml"
+    else :
+        return ".invalid"
+        
+def formatToContentType(to_format):
+    if (to_format=="TAG"):
+        return "text/tag-value"
+    elif (to_format=="RDFXML"):
+        return "application/rdf+xml"
+    elif (to_format=="XLS"):
+        return "application/vnd.ms-excel"
+    elif (to_format=="XLSX"):
+        return "application/vnd.ms-excel"
+    elif (to_format=="JSON"):
+        return "application/json"
+    elif (to_format=="YAML"):
+        return "text/yaml"
+    elif (to_format=="XML"):
+        return "application/xml"
     else :
         return ".invalid"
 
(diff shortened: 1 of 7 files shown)


goneall:
@rtgdk Looking at the unit test failures, I need to add the API support and fix the unit tests for the app views.  I'll ping you once those are updated.

goneall:
@rtgdk OK - this should be ready for review.  I updated the API code and the unit tests.

src/app/templates/app/convert.html
```diff
@@ -20,30 +20,25 @@
 		&nbsp; From &nbsp;
 			 <select name="from_format" id="from_format">
 			 	  <option value="0" name ="here" selected="">-</option>
-				  <option value="Tag">Tag</option>
-				  <option value="RDF">RDF</option>
-				  <option value="Spreadsheet">Spreadsheet</option>
+				  <option value="TAG">Tag</option>
+				  <option value="RDFXML">RDF</option>
```
rtgdk, line 24: Would we like to change the option text to RDFXML?
Screenshot 2020-10-12 at 10 32 40 PM

It shows `Upload SPDX RDFXML Document` in the upload box.

src/app/templates/app/convert.html
```diff
@@ -138,41 +133,52 @@ <h4 id="doclabel">Upload SPDX Document using the button or by dragging and dropp
...
   $("#from_format").change(function() {
-    $("#tagToRdfFormat").prop("disabled",'disabled');
-    $(".tagToRdfOptions").css("color",'grey');
     if ($(this).data('options') === undefined) {
       /*Taking an array of all options-2 and kind of embedding it on the select1*/
       $(this).data('options', $('#to_format option').clone());
     }
     var id = $(this).val();
-    var options = $(this).data('options').filter('[data-value=' + id + ']');
+	var id2 = $('#to_format').val();
```
rtgdk, line 141: nit: indent, although it doesn't matter in js/html

src/app/templates/app/convert.html
```diff
@@ -138,41 +133,52 @@ <h4 id="doclabel">Upload SPDX Document using the button or by dragging and dropp
...
     else{
       $('#doclabel').text("Upload SPDX Document using the button or by dragging and dropping onto the dashed region");
     }
-    var id2 = $('#to_format').val();
-    if (id2=="Spreadsheet"){
+	if (id2 == id) {
+	    $('#to_format').val("0");
+    } else {
+		$('#to_format').val(id2);
+	}
```
rtgdk, line 154: i think, we don't need else block, since `id2=$('#to_format').val();`

src/api/views.py
```diff
@@ -201,93 +230,24 @@ def convert(request):
...
-                        else :
-                            message, returnstatus, httpstatus = convertError('400')
+                    fromFileFormat = serFileTypeEnum.valueOf(option1);
+                    toFileFormat = serFileTypeEnum.valueOf(option2);
+                    spdxConverter.convert(uploaded_file_path,
+                                folder+sep+convertfile, fromFileFormat, toFileFormat)
+                    retval = verifyclass.verify(folder+sep+convertfile, toFileFormat)
+                    if (len(retval) > 0):
+                        warningoccurred = True
                     if (warningoccurred == True ):
```
rtgdk, line 240: We can change this to if `if (len(retval) > 0):` since warningoccurred will be True only then.

src/app/templates/app/convert.html
```diff
@@ -138,41 +133,52 @@ <h4 id="doclabel">Upload SPDX Document using the button or by dragging and dropp
...
+      $('#convertdoclabel').text(".json");
+      $('#cfileformat').val(".json");
+    }
+	else if (id2=="YAML"){
+      $('#convertdoclabel').text(".yaml");
+      $('#cfileformat').val(".yaml");
+    }
+	else if (id2=="XML"){
+      $('#convertdoclabel').text(".xml");
+      $('#cfileformat').val(".xml");
```
rtgdk, line 181: These if else can be combined by using html data property, will commit the changes.

rtgdk (commented):
Changes look great to me. Did some manual testing, works fine. leaving some minor fixes to be done. Will commit one thing from my side to improve code.

rtgdk:
@goneall Added this [commit](https://github.com/spdx/spdx-online-tools/pull/245/commits/21b84b8226a8c25f0a5a58d33aaf7a891b5ea723) to improve js code for showing the file extension.

Other than the comment for minor fixes, things look good to me. We can merge.

src/app/templates/app/convert.html
rtgdk: Fixed in my commit.

src/app/templates/app/convert.html
rtgdk: Fixed in my commit.

rtgdk (approved):

src/app/templates/app/convert.html
rtgdk: Got it, we need this since we are updating the options - `$('#to_format').html(options);` Have fixed in my new commit

rtgdk (commented):
Fixed this in [commit](https://github.com/spdx/spdx-online-tools/pull/245/commits/b35c8dfae4fa3d963846dc8e7646da8d373ee3e2)

src/app/templates/app/convert.html
goneall: Good suggestion - fixed in commit 981e0d5

src/app/templates/app/convert.html
goneall: A bit of a problem with tabs - fixed in commit 981e0d5

src/api/views.py
goneall: Fixed in commit 981e0d5

Issue #240, opened 2020-09-24
Add / modify fields for New License Request


swinslow:
Followup from #1108, filing here at @goneall's request:

From @vmbrasseur:

> The current 'new license request' issue template (by way of tools.spdx.org, I believe) has these fields:
> 
> License Name:
> Short identifier:
> License Author or steward:
> Comments:
> Standard License Header:
> URL:
> OSI Status:
> It may make sense to evolve this template a bit to include more of the information needed for the inclusion principles. For instance, maybe add a field for listing example projects using the license.
> 
> What do folks think of reconsidering the fields on this template?

I'm +1 on adding a field to list URLs to example projects using the license, this is info we almost always want.

I'll take another look through the inclusion principles to see if there's more we want to ask.

goneall:
resolved with PR #248

PR #248, opened 2020-10-25
Resolves issue #240 - update submit license fields

goneall:
Resolves issue #240 

Added project examples:
image

The example URL's are added to the issues.  Note that the project URL's are not added to the database.

For an example, see: https://github.com/spdx/TEST-LicenseList-XML/issues/92

### DIFF
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -80,6 +80,7 @@ class LicenseRequestForm(forms.Form):
     shortIdentifier = forms.CharField(label='Short identifier', max_length=25)
     sourceUrl = forms.CharField(label='Source / URL', required=False)
     osiApproved = forms.CharField(label="OSI Status", widget=forms.Select(choices=OSI_CHOICES))
+    examples = forms.CharField(label='Example Projects / URL', required=False)
     comments = forms.CharField(label='Comments', required=False, widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
     licenseHeader = forms.CharField(label='Standard License Header', widget=forms.Textarea(attrs={'rows': 3, 'cols': 40}), required=False)
     text = forms.CharField(label='Text', widget=forms.Textarea(attrs={'rows': 4, 'cols': 40}))
--- a/src/app/forms.py
+++ b/src/app/forms.py
@@ -80,7 +80,7 @@ class LicenseRequestForm(forms.Form):
     shortIdentifier = forms.CharField(label='Short identifier', max_length=25)
     sourceUrl = forms.CharField(label='Source / URL', required=False)
     osiApproved = forms.CharField(label="OSI Status", widget=forms.Select(choices=OSI_CHOICES))
-    exampleUrl = forms.CharField(label='Example Projects / URL', required=False)
+    exampleUrl = forms.CharField(label='Example Projects / URL', required=True)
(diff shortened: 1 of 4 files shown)


goneall:
@swinslow @vmbrasseur Please review the wording and approach

src/app/forms.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, *args, **kwargs):
     shortIdentifier = forms.CharField(label='Short identifier', max_length=25)
     sourceUrl = forms.CharField(label='Source / URL', required=False)
     osiApproved = forms.CharField(label="OSI Status", widget=forms.Select(choices=OSI_CHOICES))
+    examples = forms.CharField(label='Example Projects / URL', required=False)
```
rtgdk, line 83: This will cause the same issue as https://github.com/spdx/spdx-online-tools/issues/249. Since its a Char Field, so form only accept one example. In views.py
`licenseExamples = [form.cleaned_data['examples']]` will return only one example. 

We either to need to change the forms.py to accept multiple urls or send the urls as comma separated/json formatted from javascript.

Will look into it and commit the changes along with https://github.com/spdx/spdx-online-tools/issues/249 fix.

rtgdk (commented):
Changes look good to me. One major change needed - same as #249

src/app/forms.py
goneall: Agreed - any help appreciated.  I may make an attempt at a fix tomorrow based on your suggestions - I'll update the PR and issue if I'm successful.

vmbrasseur:
1. For wording, it's "URLs", not "URL's". There's no apostrophe needed for pluralisation of an acronym.
1. The code says `exampleUrl = forms.CharField(label='Example Projects / URL', required=False)`. Should that be `required=True`? I can't picture a situation where we do _not_ want to see at least one example, ideally more.

goneall:
@vmbrasseur Thanks for the review.

I just added commit https://github.com/spdx/spdx-online-tools/pull/248/commits/fcefdde9b439f3bacb33547875c1889f36f57d5d to make the field required and remove the apostrophe from URL's

goneall:
> This will cause the same issue as #249. 

@rtgdk This should be resolved with commit d2f7d09

I also fixed some related issues and an issue where the form was not properly reset on submit.

This should be good to go - if you don't mind doing one more review before merging.

src/app/forms.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, *args, **kwargs):
     shortIdentifier = forms.CharField(label='Short identifier', max_length=25)
     sourceUrl = forms.CharField(label='Source / URL', required=False)
     osiApproved = forms.CharField(label="OSI Status", widget=forms.Select(choices=OSI_CHOICES))
+    exampleUrl = forms.CharField(label='Example Projects / URL', required=True)
```
rtgdk, line 83: Should we use a TextField instead? For sourceUrl too. Although, the limit for CharField depends on the sql database we are using(255-65,535, goes to unlimited in some cases). I think we should be fine without it.
This change is not needed at all. Just pointing out for future reference.

rtgdk (approved):
Changes look good to me. We can merge this.

src/app/forms.py
goneall: I couldn't find a TextField in forms - it looks like it is only in the DB model

PR #250, opened 2020-10-31
Fix failures when submitting issues outside of license request

goneall:


goneall:
@rtgdk I missed this when doing the previous fix.  Found during some final testing.

rtgdk (approved):

PR #251, opened 2020-10-31
Add instructions for updating the production database

goneall:


goneall:
@kmehant @Ugtan - Let me know if this is the correct method for updating the database or if we should somehow include it in the Docker file creation

Ugtan (commented):
@goneall I guess this would work well with initialising or updating of the database. But, have you tried `docker-compose exec spdx_prod python manage.py migrate`?

goneall:
@Ugtan Thanks for the suggestion - The only problem I can think of is we may need to stop the running instance first.

I haven't tried using docker-compose rather than docker.

goneall:
I updated the instructions to use individual docker exec commands which is a bit easier than starting up the shell.

Issue #263, opened 2020-11-22
Uncaught ReferenceError: CodeMirror on Submit New License page


alandtse:
Visiting https://tools.spdx.org/app/submit_new_license/ results in the following JavaScript error.

Chrome:
```text
script.js:102 Uncaught ReferenceError: CodeMirror is not defined
    at HTMLDocument.<anonymous> (script.js:102)
    at i (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at Function.ready (jquery.min.js:2)
    at HTMLDocument.K (jquery.min.js:2)
```

Firefox:
```text
Uncaught ReferenceError: CodeMirror is not defined
    <anonymous> https://tools.spdx.org/static/js/editor/script.js:102
    jQuery 4
script.js:102:5
    <anonymous> https://tools.spdx.org/static/js/editor/script.js:102
    jQuery 4
```

May be related to #261.

Noticed this testing my Submit license feature in spdx-license-diff and noticing it was broken trying to inject the autofill code.

goneall:
@alandtse Thanks for reporting this.  

It turns out that this is unrelated to the deployment.  I was able to reproduce it on my local machine.

I doesn't seem to affect any of the functionality on the web page since the CodeMirror editor does not seem to be in use.

I put in PR #265  to remove the reference.

I also patched the production machine to remove the reference.

@alandtse Let me know if you see any other issues.  The API's have not gone through a lot of testing on the production environment, so let me know if you run into any issue.

goneall:
This has been resolved in PR #265 and deployed to production.

PR #264, opened 2020-11-22
Remove editor from submit

goneall:
Remove the unnecessary CodeMirror editor from the submit new license HTML file.  This does not seem to be used and causes an error to be displayed in the web browser console.

This resolves issue #263


goneall:
@techytushar @rtgdk or @Ugtan If one of you could verify my assumption that this isn't used.

goneall:
Oops - based this on the wrong branch - reopening a separate PR.

PR #205, opened 2020-07-19
Upgrade to python3

rtgdk:
Updated python3 branch with master and made changes to support python3.7+. Able to run the server, some tests are failing locally either due to bytes/str conversion or jpype upgrade.

Main changes done in this PR:

1. Updated python3 branch with master.
2. Updated requirements.txt - all packages are updated to a newer version with python3.7+ support.
3. jpype changed JavaException to JException.
4. Python3 related changes - 
      * izip changed to zip
      *  relative imports
      * Use wb instead of w when opening files - to write bytes content
5. Django2 related changes -
     * is_authenticated() changed to is_authenticated
     * on_delete=models.CASCADE added to foreign keys
     * Fixed example dir path by settings up EXAMPLES_DIR variable in setting.py
6. Ran 2to3 on the whole code
7. Updated tests to use bytes when comparison
8. Update dockerfiles, github workflow for py3
9. Updated Readme.md for py3 install and run instructions (Major change is to remove virtualenv install since python3 supports virtual env itself with `python -m venv`

Major Issue Left to Address:
1. Python3 changed how we used to represent in strings in python2. Now HttpResponse and opening/writing file uses bytes instead of string- https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit
This issue made a lot of tests fail due to comparision b/w str and bytes. Most of them are fixed by changing expected comparison to bytes by adding b in the str. Although, more robust changes might be required to come up with a way to compare actual output - either changing the response to str by encoding or str to byted by decoding. `(encode()/decode() functions of str)`

 ----- 
Some of the changes are picked from https://github.com/spdx/spdx-online-tools/pull/178, although it was not updated with master and did not have the full changes, so had to create a new PR. Thanks @victorphoenix3 for all the work.
 
----
**To review this PR, either checkout the branch locally and run the server or check the git diff with "master" instead. I'll open a new PR from python3->master once this gets merged with all failures resolved.**

Fixes #58 #24

PR #265, opened 2020-11-22
Remove unnecessary JavaScript editor inclusion

goneall:
Remove the unnecessary CodeMirror editor from the submit new license HTML file. This does not seem to be used and causes an error to be displayed in the web browser console.

This resolves issue #264 

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -652,7 +652,6 @@ async function generate_text_diff(base, newtxt){
     });
   }
 </script>
-<script type="text/javascript" src="{% static 'js/editor/script.js' %}"></script>
 <script type="text/javascript" src="{% static 'js/editor/difflib.js' %}"></script>
 <script type="text/javascript" src="{% static 'js/editor/diffview.js' %}"></script>
 <script type="text/javascript" src="{% static 'js/editor/treeview.js' %}"></script>


goneall:
@techytushar @rtgdk or @Ugtan If one of you could verify my assumption that this isn't used.

Ugtan (commented):
The script.js and treeview.js files are for the license XML editor. Hence, we can safely remove them from the license submittal. @goneall

Ugtan:
Just a suggestion, I think we should now focus more on testing the application as I feel like a lot of code is not being tested which caused a lot of hassle for you especially in production. Also, we should use code coverage to maintain test coverage for every single line of code we add to the spdx online tools. thoughts? @goneall

goneall:
> Just a suggestion, I think we should now focus more on testing the application as I feel like a lot of code is not being tested which caused a lot of hassle for you especially in production. Also, we should use code coverage to maintain test coverage for every single line of code we add to the spdx online tools. thoughts?

Agree once we complete the migration to Python3.  We have some unit tests that are failing I'd like to clean up.

rtgdk (approved):
Lgtm.

rtgdk:
@goneall @Ugtan I totally agree. We should improve our testing capabilities. I had raised a PR #205 and recently updated it with master - made all the changes for package versions update. From an initial look, all the functionality was working fine. Some test cases were failing. If you get some time, please feel free to have a look. I am planning to work more on python3 upgrade this coming weekend. We should aim for the complete upgrade by January so that any new project gets build on the new version.

goneall:
I just checked and treeview.js looks like it is used by one of the modal dialogs, so I'll go ahead and merge this.

Issue #272, opened 2020-12-10
Bytes to str conversion results in malformed xml in Python3


rtgdk:
Discussion - https://github.com/spdx/spdx-online-tools/pull/205#issuecomment-742222481

goneall:
I did some more research and the data is stored incorrectly in the database.

In the Python3 version, prepended with b:

```
>>> result[0][7]
'b\'<SPDXLicenseCollection xmlns="http://www.spdx.org/license"><license isOsiApproved="true" licenseId="testid" name="Test License [Please Ignore]" listVersionAdded=""><crossRefs><crossRef>http://source1.com</crossRef></crossRefs><standardLicenseHeader /><notes /><text><p>test not there yet</p></text></license></SPDXLicenseCollection>\''
```

I looked at the production database, and the string is just stored as XML:

```
<SPDXLicenseCollection xmlns="http://www.spdx.org/license"><license isOsiApproved="false" licenseId="V8" listVersionAdded="" name="V8 Project License"><crossRefs /><standardLicenseHeader /><notes /><text><p>Copyright 2006-2011, the V8 project authors. All rights reserved.
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are
met:</p><list><list><item><bullet>*</bullet><p>Redistributions of source code must retain the above copyright
      notice, this list of conditions and the following disclaimer.
     Redistributions in binary form must reproduce the above
      copyright notice, this list of conditions and the following
      disclaimer in the documentation and/or other materials provided
      with the distribution.
     Neither the name of Google Inc. nor the names of its
      contributors may be used to endorse or promote products derived
      from this software without specific prior written permission.</p><p>THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
[...]

PR #273, opened 2020-12-10
Fix malformed xml  which converting bytes to str in py3

rtgdk:
#272

### DIFF
--- a/src/api/oauth.py
+++ b/src/api/oauth.py
@@ -24,7 +24,9 @@ def generate_github_access_token(github_client_id, github_client_secret, github_
         }),
         headers={'content-type': 'application/json'}
     )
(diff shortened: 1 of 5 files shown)


goneall:
I tried this out on the server that has the bad data and I got the following error:

`AttributeError: 'str' object has no attribute 'decode'`

I look in the database and the data is stored as a quoted string prepended with b (e.g. `b'<xml stuff`).

rtgdk:
Oh 😓 Then I believe we need to look through the codebase for other model fields too since some might be saved as bytes and some as string. To fix this we can add the check to do decode if it is not an instance of string.

goneall:
This is now working on my test machine, but there are two new issues: #275 and #274

goneall:
I just created a PR #276 against this PR branch to resolve the issues mentioned above

goneall (approved):

Issue #275, opened 2020-12-16
In Python3 - license text is displayed as a byte type


goneall:
When selecting to view a license request from the license requests table, the text shows prepended with `b'` and the quotes are escaped.

The full display is below:

```
Test License [Please Ignore]

Fullname

Test License [Please Ignore]

License author name

Short Identifier

testid

Source/URL

http://source1.com

Osi Approved

true

Notes

None

Standard License Header

None

Text

b'

this doesn\'t match

'
Submission Date

Dec. 16, 2020, 11:12 p.m.
```

goneall:
Resolved in PR #276

goneall:
Resolved with PR #276

Issue #274, opened 2020-12-16
In Python3 - error writing XML to file


goneall:
When downloading the xml file for a license submittal, the following error occurs:

```
  File "C:\Users\gary\git\temp\spdx-online-tools\venv\lib\site-packages\django\core\handlers\exception.py", line 34, in inner
    response = get_response(request)
  File "C:\Users\gary\git\temp\spdx-online-tools\venv\lib\site-packages\django\core\handlers\base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "C:\Users\gary\git\temp\spdx-online-tools\venv\lib\site-packages\django\core\handlers\base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Users\gary\git\temp\spdx-online-tools\src\app\views.py", line 360, in licenseInformation
    xmlFile.write(xmlString)
TypeError: a bytes-like object is required, not 'str'
```

jlovejoy:
is this still a unique open issue? seems similar to something more recent (or has been fixed?)

@goneall @rtgdk

PR #276, opened 2020-12-17
Fix encoding issues for open and XML parsing

goneall:
Resolves issues #275 and #274


rtgdk (approved):
LGTM. Added some more similar changes.

PR #291, opened 2021-02-12
Fix #241: Added automatic filename in convert

### LINKED ISSUE #241: [Suggestion] Automatic Output File Name
While testing [/app/convert/](http://52.32.53.255/app/convert) converter, I was required to enter the name for the converted file. It would've been better if the application auto-set the converted file's basename the same as that of the input file.
For example: if "input.rdf" is the input file that is to be converted to xls format, the converted file name could've been set to "input" as a placeholder.

hardikkat24:
Fixes #241
This PR adds automatic filename in url `app/convert/` according to the file entered after removing the extension of the file.
Example: `test.extension` will have default Converted File Name as `test`

### DIFF
--- a/src/app/templates/app/convert.html
+++ b/src/app/templates/app/convert.html
@@ -120,7 +120,7 @@
   function checkFile(file){
     if(file){
       $(".file-name").html("<b>Selected File: </b>"+file.name);
-      $("#cfilename").val(file.name.split(".")[0]);
+      $("#cfilename").val(file.name.split('.').slice(0,-1).join('.'));
       selected_file = file;
     }
     else{
--- a/src/app/templates/app/convert.html
+++ b/src/app/templates/app/convert.html
@@ -120,6 +120,7 @@
   function checkFile(file){
     if(file){
       $(".file-name").html("<b>Selected File: </b>"+file.name);
+      $("#cfilename").val(file.name.split(".")[0]);
       selected_file = file;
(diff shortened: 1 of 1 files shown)


rtgdk:
We would need similar changes elsewhere like in compare.
Also, can we handle this any better? This would not work correctly for filename like `test.filename.extension`.

hardikkat24:
@rtgdk I have made changes so that filenames like `name.xxx.yyy.ext` can be handled correctly.
Now compare takes multiple files, so what should I keep the name of the file?
Thanks!

hardikkat24:
@rtgdk Please take a look.

RishabhBhatnagar:
@hardikkat24 thank you for your PR!

There's one bug I can see. What if the filename doesn't have an extension? (this might not happen but is a valid case that our system allows).
The current implementation will trim the name of the file. For example, if the file name is `spdx_document`, output will be an empty string.
Also, please signoff your commits as mentioned in the [Contributing guidelines](https://github.com/spdx/spdx-online-tools/blob/master/CONTRIBUTING.md)

hardikkat24:
@RishabhBhatnagar Thanks a lot for your feedback. I forgot to test my code for this case. I have made changes accordingly. Please take a look.

rtgdk (approved):
@hardikkat24 Thanks for the contribution!

PR #305, opened 2021-04-04
Remove depreciated GoogleOpenId and YahooOpenId

nishantc7:
Authentication Backends for GoogleOpenId and YahooOpenId are [depreciated](https://github.com/python-social-auth/social-core/issues/472) in [python-social-auth](https://github.com/python-social-auth/social-core/blob/master/CHANGELOG.md) 4.0 . Thus removing these fixes module load error in `app.tests.ProfileViewsTestCase` .

### DIFF
--- a/src/src/settings.py
+++ b/src/src/settings.py
@@ -144,10 +144,6 @@ AUTH_PASSWORD_VALIDATORS = [
 ]
 
 AUTHENTICATION_BACKENDS = [
-    'social_core.backends.open_id.OpenIdAuth',
-    'social_core.backends.google.GoogleOAuth2',
-    'social_core.backends.google.GoogleOAuth',
-    'social_core.backends.twitter.TwitterOAuth',
     'social_core.backends.github.GithubOAuth2',
     'django.contrib.auth.backends.ModelBackend',
     'rest_framework_social_oauth2.backends.DjangoOAuth2',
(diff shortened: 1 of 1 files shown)


src/src/settings.py
```diff
@@ -145,11 +145,9 @@
...
 AUTHENTICATION_BACKENDS = [
     'social_core.backends.open_id.OpenIdAuth',
-    'social_core.backends.google.GoogleOpenId',
     'social_core.backends.google.GoogleOAuth2',
     'social_core.backends.google.GoogleOAuth',
     'social_core.backends.twitter.TwitterOAuth',
-    'social_core.backends.yahoo.YahooOpenId',
     'social_core.backends.github.GithubOAuth2',
     'django.contrib.auth.backends.ModelBackend',
     'rest_framework_social_oauth2.backends.DjangoOAuth2',
```
rtgdk, line 153: Thanks for the PR. Do we need to use other auth backends? We only use Github right now. Login with other medium is blocked I think. So is it better to remove all these dependencies and clear out code in python3?

rtgdk (approved):

src/src/settings.py
nishantc7: Yes, I agree. We could remove the additional Twitter and Google backends, in case we have no plans to provide these logins in the future.

rtgdk: Yeah lets remove unnecessary auth backends

nishantc7: All done, working as expected.

rtgdk (approved):

PR #316, opened 2021-06-22
Merge App and Api by creating a common utility for core logic

### LINKED ISSUE #287: Merge app and api code under a common utils
Right now, api `views.py` code is not updated with the changes in app `views.py`. There are some API which doesn't use anything from app source code. Not sure why, but it was a bad design decision to separate API and App code. We should have a common utility handling both API and app calls.

Ugtan:
Fixes #287 

The PR adds support for a common utility that contains all the core logic required for the SPDX Online tools. Both App and Api make use of this utility so, we don't have to test the app and api individually. It reduces redundancy and makes it easy to test the modules and maintain them.

### DIFF
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -37,6 +37,10 @@ import re
 import datetime
 import xml.etree.cElementTree as ET
 import app.core as core
+import api.utils as utils
+
+from app.generateXml import generateLicenseXml
+from app.utils import createIssue
 
 from traceback import format_exc
 from os.path import abspath, join, sep, splitext
@@ -72,6 +76,7 @@ class CompareViewSet(ModelViewSet):
     serializer_class = CompareSerializerReturn
     parser_classes = (MultiPartParser, FormParser,)
 
+
 @api_view(['GET', 'POST'])
 @renderer_classes((JSONRenderer,))
 def validate(request):
@@ -88,28 +93,10 @@ def validate(request):
         
         if serializer.is_valid():
             core.initialise_jpype()
-            output = core.license_validate_helper(request)
+            response = core.license_validate_helper(request)
+            httpstatus, _, result = utils.get_json_response_data(response)
             jpype.detachThreadFromJVM()
-
-            httpstatus = output.get('status', None)
-            context_dict = output.get('context', None)
-            response = output.get('response', None)
-            message = output.get('message', None)
-
-            if context_dict:
-                result = context_dict.get('error', None)
-            elif response:
-                result = response
-            else:
-                result = message
-            
-            if httpstatus == 200:
-                returnstatus = status.HTTP_200_OK
-            elif httpstatus == 400:
-                returnstatus = status.HTTP_400_BAD_REQUEST
-            else:
-                returnstatus = status.HTTP_404_NOT_FOUND
-            
+            returnstatus = utils.get_return_code(httpstatus)
             query = ValidateFileUpload.objects.create(owner=request.user, file=request.data.get('file'))
             query.result = result
             query.status = httpstatus
@@ -125,47 +112,6 @@ def validate(request):
                 serializer.errors, status=status.HTTP_400_BAD_REQUEST
                 )
 
-def extensionGiven(filename):
-    if (filename.find(".")!=-1):
-        return True
-    else:
-        return False
-
-def getFileFormat(to_format):
-    if (to_format=="TAG"):
-        return ".spdx"
-    elif (to_format=="RDFXML"):
-        return ".rdf.xml"
-    elif (to_format=="XLS"):
-        return ".xls"
-    elif (to_format=="XLSX"):
-        return ".xlsx"
-    elif (to_format=="JSON"):
-        return ".json"
-    elif (to_format=="YAML"):
-        return ".yaml"
-    elif (to_format=="XML"):
-        return ".xml"
-    else :
-        return ".invalid"
-        
-def formatToContentType(to_format):
-    if (to_format=="TAG"):
-        return "text/tag-value"
-    elif (to_format=="RDFXML"):
-        return "application/rdf+xml"
-    elif (to_format=="XLS"):
-        return "application/vnd.ms-excel"
-    elif (to_format=="XLSX"):
-        return "application/vnd.ms-excel"
-    elif (to_format=="JSON"):
-        return "application/json"
-    elif (to_format=="YAML"):
-        return "text/yaml"
-    elif (to_format=="XML"):
-        return "application/xml"
-    else :
-        return ".invalid"
 
 @api_view(['GET', 'POST'])
 @renderer_classes((JSONRenderer,))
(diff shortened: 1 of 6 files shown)


src/api/views.py
```diff
@@ -20,6 +20,7 @@
 from api.serializers import ValidateSerializer,ConvertSerializer,CompareSerializer,CheckLicenseSerializer,SubmitLicenseSerializer,ValidateSerializerReturn,ConvertSerializerReturn,CompareSerializerReturn,CheckLicenseSerializerReturn,SubmitLicenseSerializerReturn
```
rtgdk, line 20: Don't force push commits. Makes it hard to review changes. We can squash commits while merging

src/api/views.py
```diff
@@ -83,58 +84,37 @@ def validate(request):
...
-                file=request.data.get('file')
-            )
+            initialise_jpype()
+            output = license_validate_helper(request)
+            jpype.detachThreadFromJVM()
+
+            httpstatus = output.get('status', None)
+            context_dict = output.get('context', None)
+            response = output.get('response', None)
+            message = output.get('message', None)
```
rtgdk, line 96: Have a common function for this too, as this conversion if needed in every api

src/api/views.py
```diff
@@ -83,58 +84,37 @@ def validate(request):
...
+                result = response
+            else:
+                result = message
+            
+            if httpstatus == 200:
+                returnstatus = status.HTTP_200_OK
+            elif httpstatus == 400:
+                returnstatus = status.HTTP_400_BAD_REQUEST
+            else:
+                returnstatus = status.HTTP_404_NOT_FOUND
```
rtgdk, line 110: Common function for this conversion too

src/api/views.py
```diff
@@ -83,58 +84,37 @@ def validate(request):
...
+            context_dict = output.get('context', None)
+            response = output.get('response', None)
+            message = output.get('message', None)
+
+            if context_dict:
+                result = context_dict.get('error', None)
+            elif response:
+                result = response
+            else:
+                result = message
```
rtgdk, line 103: Can we simplify this logic?

src/api/views.py
```diff
@@ -83,58 +84,37 @@ def validate(request):
...
-                returnstatus = status.HTTP_400_BAD_REQUEST
-                httpstatus = 400
-                jpype.detachThreadFromJVM()
-            query.result=result
-            query.status=httpstatus
-            ValidateFileUpload.objects.filter(file=uploaded_file).update(result=result,status=httpstatus)
+            ValidateFileUpload.objects.filter(file=uploaded_file).update(result=result, status=httpstatus)
             serial = ValidateSerializerReturn(instance=query)
             return Response(
                 serial.data, status=returnstatus
```
rtgdk, line 120: Can we other functions apart from views functions like `extensionGiven` , `getFileFormat`, etc. to a utils.py file? Basically this file should contain only core logic, other business logic function should be moved to utils function

src/app/core.py
```diff
@@ -0,0 +1,468 @@
...
+    if not jpype.isJVMStarted():
+        classpath = settings.JAR_ABSOLUTE_PATH
+        jpype.startJVM(jpype.getDefaultJVMPath(), "-ea", "-Djava.class.path=%s"%classpath)
+            
+    # Attach a thread to JVM and start processing
+    jpype.attachThreadToJVM()
+
+
+def license_compare_helper(request):
+    """
```
rtgdk, line 31: Just confirming that you copied this code from app/views.py, right? Coz similar api code was outdated.

src/api/views.py
```diff
@@ -20,6 +20,7 @@
 from api.serializers import ValidateSerializer,ConvertSerializer,CompareSerializer,CheckLicenseSerializer,SubmitLicenseSerializer,ValidateSerializerReturn,ConvertSerializerReturn,CompareSerializerReturn,CheckLicenseSerializerReturn,SubmitLicenseSerializerReturn
 from api.oauth import generate_github_access_token,convert_to_auth_token,get_user_from_token
 from app.models import LicenseRequest
+from app.core import initialise_jpype, license_validate_helper, license_check_helper
```
rtgdk, line 23: Import core instead of specific functions. Make it hard to maintain and also better understanding of code flow.

rtgdk (changes requested):
Can you post the test case status for these changes?

rtgdk:
@Ugtan Looks good in the first iteration. Can you also attach test case run results here after every PR? Just want to make sure we are not opening a new hole in our application.

src/api/views.py
Ugtan: Alright. I won't force push commits from now onwards.

src/api/views.py
Ugtan: Sure. I was thinking the same. It would reduce code redundancy as this piece of code is used both in app and api.

src/api/views.py
Ugtan: Alright.

src/api/views.py
Ugtan: Sure. I will do that.

src/app/core.py
Ugtan: Yeah, it's the code from app views.

Ugtan:
> @Ugtan Looks good in the first iteration. Can you also attach test case run results here after every PR? Just want to make sure we are not opening a new hole in our application.

@rtgdk Sure, I will attach a screenshot of the API test cases once I complete the other two APIs for compare and convert most probably today itself. And, I'm testing both APP and API to make sure both of them work simultaneously.

Ugtan:
@rtgdk Could you take a look at it once, please? @SantiagoTorres Would definitely appreciate your review as well. thanks

I have merged App and Api code for license Validate, Compare, Convert, and Check License. I will also attach the screenshot of the test cases of all these functions below.

For Validate API
ValidateApiTests

For Convert API
ConvertApiTests

For Compare API
compareApiTests

For Check License API(Not passing idk why)
image

Getting this re-module error not sure why. Will try to fix it.
image

src/app/core.py
```diff
@@ -0,0 +1,469 @@
...
+
+from django.utils.datastructures import MultiValueDictKeyError
+from django.core.files.storage import FileSystemStorage
+from django.conf import settings
+
+from json import dumps, loads
+from time import time
+from traceback import format_exc
+from urllib.parse import urljoin
+from app.views import formatToContentType, getFileFormat
```
rtgdk, line 14: Move these functions to utils.py, views.py should just be a wrapper over helper functions.

rtgdk, line 14: This will be a cyclic dependency kind of logic, views depends on core, core depends on views for some functions

rtgdk (commented):
Great work @Ugtan left some comments to fix. I'll it merge it post that.

rtgdk:
@Ugtan For check license api, it might be failing due to python3 str/bytes issue on spdx_license_matcher side.

rtgdk:
@Ugtan Merged the django3 PR. Please rebase those changes with this PR and post the test results again for both app and api. Also, please make sure that we didn't lose any functions/functionality due to these changes. I'll merge after you post the results post rebase.

src/app/core.py
Ugtan: Alright, I will move these functions to utils.py file

Ugtan: Understood.

Ugtan:
> For check license api, it might be failing due to python3 str/bytes issue on spdx_license_matcher side.

I'm not quite sure regarding the issue as it works perfectly when I use the check license functionality on localhost. I also tried to run the SPDX License Matcher and it worked as well. I'm facing the issue only while running the test cases to check license API.


> @Ugtan Merged the django3 PR. Please rebase those changes with this PR and post the test results again for both app and api. Also, please make sure that we didn't lose any functions/functionality due to these changes. I'll merge after you post the results post rebase.

I will rebase the branch and share the test results for both the app and API with you. I have briefly tested everything and the test cases also working fine so I don't think we will run into any problem.

rtgdk:
@Ugtan Can you share the test results?

Ugtan:
Here you go @rtgdk 

Test results for the API:

image

image

image

image


The APP test cases are failing after the changes I'm not quite sure why 15 of them failed with the following error:

image
image


I'm not quite sure why are these test cases failing. I also tried to dig a little but couldn't figure out the root cause for the error to occur. I have been trying for some time, Could you help if possible? @rtgdk

Ugtan:
@rtgdk Fixed it. Both APP and API test cases are working. I have already shared the API ones. The test results for the app:

image

rtgdk (approved):
@Ugtan Thanks! Merging this.

Issue #306, opened 2021-04-09
Extremely large SPDX files times out


goneall:
Suggest a short term solution where we have a size limit on SPDX documents for validate, translate, and convert then produce an error and suggest they use the command line tool available at github.com/spdx/tools-java

Longer term, we could do some kind of batch processing approach and not suffer the processing timeouts.

goneall:
@dkruszew reported the same issue and included the attached file as a reference.

In looking at the file, it is quite large at 122,244 lines.

As a work-around, the latest commandline version of the SPDX online tools from the [release page](https://github.com/spdx/tools-java/releases/tag/v1.0.1) and run locally.

I tried this and it took 1 minute 46 seconds on a laptop.  This is processing 1,153 lines per second - slower that I would of expected but not crazy unreasonable.

I noticed you have a lot of relationships in this document.  One optimization that would make a potentially big difference is to cache the verification results of each element so we don't end up re-traversing the same nodes when we reference the same file.

I've added an issue to track this enhancment: https://github.com/spdx/tools-java/issues/20
[keycloak-13.0.1.zip-too-big.txt](https://github.com/spdx/spdx-online-tools/files/6712837/keycloak-13.0.1.zip-too-big.txt)

goneall:
Resolved with PR #318 and PR #319

PR #318, opened 2021-06-29
Add warning for large files in validate/conversion/convert

rtgdk:
#306

### DIFF
--- a/src/app/static/js/warning.js
+++ b/src/app/static/js/warning.js
@@ -1,0 +1,31 @@
+let MAX_FILE_SIZE = 10485760; // 10 MB in bytes
+let MAX_FILE_SIZE_STR = "10MB";
+function checkFileSize(file, warningElementJQ) {
+    if (file && file.size > MAX_FILE_SIZE) {
+        $(warningElementJQ).prop("hidden", false)
+            .html("WARNING: The file size exceeds the max supported file size " + MAX_FILE_SIZE_STR +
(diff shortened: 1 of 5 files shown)


rtgdk:
Screenshot 2021-06-29 at 4 00 21 PM
@goneall Let me know what to change the file size limit to and warning msg

src/app/static/js/warning.js
```diff
@@ -0,0 +1,31 @@
+let MAX_FILE_SIZE = 10485760; // 10 MB in bytes
+let MAX_FILE_SIZE_STR = "10MB";
+function checkFileSize(file, warningElementJQ) {
```
goneall, line 3: Looking at issue #306 and the attached file, 6MB will cause issues.  Let's go with 5MB as the size for a warning.

goneall (changes requested):
Thanks @rtgdk - just one change request to update the size.  Otherwise, LGTM.

src/app/static/js/warning.js
rtgdk: Fixed it 👍

goneall (approved):
Thanks @rtgdk - I'll go ahead and merge

goneall:
@kestewart This is the PR we talked about on the tech call that provides better feedback for large files.

I'll merge the PR, but I'm going to wait to deploy to see if there are other tooling fixes based on the NTIA plugfest.

PR #323, opened 2021-08-11
Add support for API documentation

Ugtan:
This PR adds support for API documentation with the help of drf-yasg library.

PR #324, opened 2021-08-15
Add diff section tool

### LINKED ISSUE #186: Add a License Diff section to the SPDX Online Tools
We can add a new section to the Online tools which will take license text from the user and shows diff if the license entered by the user is a close match. That way we could inform the user beforehand that a license that closely matches the license user has entered is already present on the SPDX License List. Alternatively, we can improve the check license functionality to display license IDs for perfect matches and display diff in case of close matches. Currently the only way to check for close matches is if the user submits the license request and then, a pop up generates.

Ugtan:
Fixes #186 

The PR adds support for a new tool called license diff.

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -1,5 +1,8 @@
 """This file contains the core logic used in the SPDX Online Tools' APP and API"""
 
+import json
+
+from django.http.response import JsonResponse
 import jpype
 import os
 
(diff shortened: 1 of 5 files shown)


rtgdk:
@Ugtan So this is a new tool? Weren't we planning to add this api to submit license page also?

rtgdk:
@Ugtan Let's update the PR and close this.

rtgdk:
As mentioned on https://github.com/spdx/spdx-online-tools/pull/323, not sure if the generated swagger specs are correct or are we missing anything, closing this one.

rtgdk:
Continued work on this with https://github.com/spdx/spdx-online-tools/pull/324

rtgdk:
Merged the changes with #345 to python3 branch. Thanks @Ugtan for the work.

PR #348, opened 2022-04-03
Resolve database error when installing or running unit tests on Python3 branch

goneall:
Remove the file `src/api/migrations/0002_submitlicensemodel.py`

This seems to solve the unit test failures on the Python3 branch.

I don't know if removing this file causes any other side effects.


goneall:
Oops - wrong branch!

PR #358, opened 2022-04-11
Disable license diff feature

### LINKED ISSUE #355: License Diff Issues
The new License-Diff feature has a couple of issues:

- It is similar to license-check functionality
- If you have a close match (e.g. just remove one or two words), it shows an exact match
- If you remove too many words from the license, it statues no match to any license

Suggest we remove this menu item / feature until the above issues are resolved.

rtgdk:
Resolves #355 

Disables newly added license diff feature by limiting access from UI, commented out it's usecases


goneall (approved):
LGTM

PR #379, opened 2022-07-01
removed "Email address" field

jewishmoses:
As asked here: https://github.com/spdx/spdx-online-tools/issues/367

PR #320, opened 2021-07-21
Add the option to raise issue when license submission failed

Ugtan:
Give an option to the license submitter to raise an issue to the SPDX online tools repository when an error has been encountered. This will make it easier to track the errors and fix them asap.

PR #384, opened 2022-08-13
Update to support SPDX Spec version 2.3

goneall:


goneall:
@rtgdk I'm going to update the production with the updated library for the new SPDX release 2.3

Let me know if we should include any other fixes or PR's.

rtgdk:
@goneall I don't think we have any other PRs fully completed. 
#323 does not provide any value.
#320 needs some confirmations before merging. Haven't recieved any reply from the author on that. Will to pick this up in my free time.
#379 needs backend changes.

Other open PRs doesn't provide any real value right now.

goneall:
@rtgdk Thanks for the reviews.  I may do another release in about 3 weeks depending on if we find any issues with the 2.3 changes.

PR #387, opened 2022-08-31
Update submit_new_license.html

jlovejoy:
Update headings as per SPDX-legal discussion on Aug 25th

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -141,15 +141,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
         </div>
       </div>
     </div>
-
-    <div class = "form-group">
-      <div class="col-sm-12">
-      <label class="control-label col-sm-3" >Email address</label>
-        <div class="col-sm-4">
-          {{ form.userEmail }}
-        </div>
-      </div>
-    </div>
 	
 	<div class = "form-group">
       <div class="col-sm-12">
@@ -441,7 +432,6 @@ function scrollUpAndHighlight(element) {
                 $("#osiApproved").val(0);
                 $("#comments").val("");
                 $("#text").val("");
-                $("#userEmail").val("");
                 var sources = $('[id="addedSourceUrl"]');
                 if(sources != null){
                   for (var i = 0; i < sources.length; i++) {
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -74,7 +74,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
         <a href="https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md">
           license inclusion principles
         </a>
-        <span class="glyphicon glyphicon-question-sign" id="commentsInfo" data-toggle="popover" data-placement="auto left"></span>
       </label>
         <div class="col-sm-6">
           {{ form.comments }}
@@ -85,7 +84,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
     <div class = "form-group">
       <div class="col-sm-12">
       <label class="control-label col-sm-3" >Full name of license (if applicable) <br /> Enter known title. If none, suggest one or leave blank
-        <span class="glyphicon glyphicon-question-sign" id="fullnameInfo" data-toggle="popover" data-placement="auto left"></span>
       </label>
         <div class="col-sm-4">
           {{ form.fullname }}
@@ -107,7 +105,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
     <div class = "form-group">
       <div class="col-sm-12">
       <label class="control-label col-sm-3" >URL to license text
-        <span class="glyphicon glyphicon-question-sign" id="sourceUrlInfo" data-toggle="popover" data-placement="auto left"></span>
       </label>
         <div class="col-sm-4">
           {{ form.sourceUrl }}
@@ -139,7 +136,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
     <div class = "form-group">
       <div class="col-sm-12">
       <label class="control-label col-sm-3" >Text of license
-        <span class="glyphicon glyphicon-question-sign" id="textInfo" data-toggle="popover" data-placement="auto left"></span>
       </label>
         <div class="col-sm-6">
           {{ form.text }}
@@ -170,7 +166,6 @@ Once you have submitted a license, please follow the <a href="https://github.com
 	<div class = "form-group">
       <div class="col-sm-12">
       <label class="control-label col-sm-3" >URLs to project that use license
-        <span class="glyphicon glyphicon-question-sign" id="exampleInfo" data-toggle="popover" data-placement="auto left"></span>
       </label>
 		<div class="col-sm-4">
           {{ form.exampleUrl }}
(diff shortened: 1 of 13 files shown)


goneall (approved):
@jlovejoy - Thanks for pulling this together - you're now an official DJango programmer ;)

Were you able to test these out on your local machine?  There are 2 ways to setup a local environment to run this on your local computer.  That way, you can check out it out visually before it is fully deployed.

jlovejoy:
I have not and I have no idea as to how to do that!

Also, there are some more changes that I think we might want to make. Unless you think it's better to do stuff more incrementally?

goneall:
> Also, there are some more changes that I think we might want to make. Unless you think it's better to do stuff more incrementally?

It would be nice to get all the changes into one branch or one PR and then test it out.  We could setup some time and I can test it on a zoom call and screen share it with you and anyone else who's interested.  I just want to make sure everything looks good before it is deployed.  It takes me about an hour to deploy, and only a few minutes to update the UI when it's running on my local machine.

Ideally, we could get you setup to test locally on your machine, but we could do that later.

jlovejoy:
@goneall @rtgdk  - now that #389 is merged.. do I need to re-do this one since it deals with the same file??

rtgdk:
@jlovejoy No need to redo if there are no changes on the same line number as #389. If changes are on different line no. then git will handle it automatically.

I would suggest to `pull/rebase` latest master into this branch and test that the page looks like just what you wanted. Overall, changes LGTM and we can merge on confirmation.

goneall:
@jlovejoy - Let me know if you'd like any help setting up a local environment for testing.

src/app/templates/app/submit_new_license.html
```diff
@@ -22,11 +22,11 @@
...
       <div class="form-group">
           <label for="branchName">Branch Name:</label>
-          <span class="glyphicon glyphicon-info-sign" data-toggle="tooltip" title="Pull Request will be made from this branch to master branch" style="font-size:12px;"></span>
+          <span class="glyphicon glyphicon-info-sign" data-toggle="tooltip" title="Pull Request will be made from this branch to main branch" style="font-size:12px;"></span>
           <input type="text" class="form-control" id="branchName" name="branchName">
       </div><br>
       <div class="checkbox">
-          <label><input type="checkbox" id="updateUpstream"> Update with upstream master </label> <span class="glyphicon glyphicon-info-sign" data-toggle="tooltip" title="Select this if you want to update your forked repo with the upstream repo, before making the pull request" style="font-size:12px;"></span>
+          <label><input type="checkbox" id="updateUpstream"> Update with upstream main </label> <span class="glyphicon glyphicon-info-sign" data-toggle="tooltip" title="Select this if you want to update your forked repo with the upstream repo, before making the pull request" style="font-size:12px;"></span>
```
jlovejoy, line 29: do we want the hover over items in this window?

src/app/templates/app/submit_new_license.html
```diff
@@ -151,7 +135,7 @@
```
jlovejoy, line 135: Check License Diff is commented out - why? (did this ever work, I can't remember?)

jlovejoy:
> @jlovejoy - Let me know if you'd like any help setting up a local environment for testing.

I think I'm ready! and have a couple other questions/comments for you to discuss!

jlovejoy:
note as per Oct 27th legal call:
- also remove hover-overs in Github PR dialog box for XML editor
- question on branch field - what is supposed to go here? (can we simplify)

src/app/templates/app/submit_new_license.html
rtgdk: Yes, this was disabled in April this year following this issue - https://github.com/spdx/spdx-online-tools/issues/355 

We can take a call to re-enable this on this page after this gets in https://github.com/spdx/spdx-online-tools/pull/392 which improves matching

src/app/templates/app/submit_new_license.html
rtgdk: I think we can remove hover from here as well and keep all help text in the same text

rtgdk (approved):
LGTM

goneall:
@jlovejoy I merged in the recent changes to the master branch which should fix the problems you were running into with docker - just pull the latest and it should work

src/app/templates/app/submit_new_license.html
jlovejoy: I removed these in last commit

jlovejoy:
@goneall @rtgdk - in the Github PR dialog box - do either of you know why there is a "commit message" field and a "PR body text" field? Don't these end up being the same thing?

jlovejoy:
Okay, figured out how to test this and made some updates to the text describing the fields. I'm still debating the short identifier field which currently says:
"Suggest short identifier (consisting of: ASCII letters (A-Z, a-z), digits (0-9), full stops (.) and hyphens (-). No spaces."

Which is kind of long and clunky. What if instead it simply said:

"Suggest short identifier according to the guidance in section B (link here: https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md )

@swinslow - thoughts?

jlovejoy:
I dont think I can test the Github PR dialog though.

goneall:
> in the Github PR dialog box - do either of you know why there is a "commit message" field and a "PR body text" field? Don't these end up being the same thing?

@jlovejoy There are actually different fields in Github.  Most of the time they are the same, but sometimes people may want them to differ.  There is a different standard for what the content and length of these fields should be - but for our purposes, they will most likely be the same.

I have no problem replacing both with a "comment and PR" field, but I also don't think it is too bad the way it is since they are different fields and some people may complain if we combine them.

One cool way to handle it would be to automatically populate the send field with the content of the first - which I think can be done in JavaScript - but I won't have time to code this up - very busy this week.

goneall:
> I dont think I can test the Github PR dialog though.

When you're all done, I can test this part out before pushing to production (I just have to remember to do it!).

swinslow:
> Okay, figured out how to test this and made some updates to the text describing the fields. I'm still debating the short identifier field which currently says:
> "Suggest short identifier (consisting of: ASCII letters (A-Z, a-z), digits (0-9), full stops (.) and hyphens (-). No spaces."
> 
> Which is kind of long and clunky. What if instead it simply said:
> 
> "Suggest short identifier according to the guidance in section B (link here: https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md )

@jlovejoy I'm kind of indifferent here, to be honest? The link to section B is more complete, though it also requires the requester to go there and read the guidance. I'm fine with either since ultimately this is just a suggestion by the submitter, and the Legal Team will review / revise as we feel appropriate in any case.

jlovejoy:
ok, @goneall - I think this is ready to go from my end!

goneall:
@jlovejoy @rtgdk I just tested this out on my local machine and I'm not able to submit.

The reason is the call to `form.is_valid()` is returning false.

goneall:
Found the issue - we removed email from the JavaScript, but it is still required on the back-end.

I just pushed a commit which fixes this

goneall:
I noticed another issue when reviewing the submitted issue.  It includes the field Standard License Header even though it was removed from the UI.

@jlovejoy Is this OK?  If so, I'll go ahead and merge.

```
1. License Name: Test license - please ignore
2. Short identifier: test-license
3. License Author or steward: Unknown
4. Comments: Test license - please ignore
5. Standard License Header:
6. License Request Url: http://localhost:8000/app/license_requests/4
7. URL(s): http://example.com/
8. OSI Status: Not Submitted
9. Example Projects: http://example1.com/

```

jlovejoy:
Thanks @goneall and sorry for the extra work.

As for the StandardHeader - I think all references to that can be removed, if possible - sounds like I missed something, if I understand correctly?

goneall:
> As for the StandardHeader - I think all references to that can be removed, if possible - sounds like I missed something, if I understand correctly?

@jlovejoy yes - it the email sending code, we just need to remove the line and probably change the numbering.  It's probably in the file views.py.

goneall:
@jlovejoy  - since I added a commit, be sure to do a git pull before making any changes

PR #389, opened 2022-08-31
remove userEmail from license request tool

jlovejoy:
accompanying changes to #379 

### DIFF
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -110,7 +110,7 @@ def submitNewLicense(request):
                 if form.is_valid() and request.is_ajax():
                     licenseAuthorName = form.cleaned_data['licenseAuthorName']
                     licenseName = form.cleaned_data['fullname']
-                    licenseIdentifier = form.cleaned_data['shortIdentifier
+                    licenseIdentifier = form.cleaned_data['shortIdentifier']
                     licenseOsi = form.cleaned_data['osiApproved']
                     licenseSourceUrls = request.POST.getlist('sourceUrl')
                     licenseExamples = request.POST.getlist('exampleUrl')
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -146,7 +146,7 @@ def submitNewLicense(request):
                             data['licenseHeader'] = licenseHeader
                             data['licenseNotes'] = licenseNotes
                             data['licenseAuthorName'] = licenseAuthorName
-                            data['userEmail'] = userEmail
+                        
                             data['comments'] = licenseComments
                         data['matchIds'] = matchingIds
                         statusCode = 409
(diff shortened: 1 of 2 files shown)


jlovejoy:
does "userEmail" also need to be removed on line 150?

goneall (approved):
@jlovejoy - Thanks for pulling this together - you're now an official DJango programmer ;)

Were you able to test these out on your local machine?  There are 2 ways to setup a local environment to run this on your local computer.  That way, you can check out it out visually before it is fully deployed.

goneall (changes requested):
@jlovejoy Please ignore my last comment - that was intended for your other PR (clicked on the wrong tab).

Could you merge in PR #379 into this PR?  That way we can test if it works before merging.

jlovejoy:
@goneall how do I merge #379 into this PR? Can I do that from the web interface? this seems like an advanced move...

goneall:
> how do I merge https://github.com/spdx/spdx-online-tools/pull/379 into this PR? Can I do that from the web interface? this seems like an advanced move...

Hmm - don't know about the web interface.  I only use that for the minutes.  From the command line, you would setup a remote for the branch the other PR was using then merge from that branch - a bit of git magic involved.  You could also just copy/paste the changes if they are small enough.  BTW - You're already in the advanced class doing some Django / Python coding ;)

jlovejoy:
@goneall - I can just add the other changes here, that's easier. Speaking of which/in that case - would it be better to simply add all of the remove-email address field changes to #387 ?

rtgdk:
@jlovejoy Yes, you can add this patch branch on top of the patch branch of #387, just need to change the base branch of this PR from `master` to desired one

I have merged #379 into this PR by changing the base branch to `jlovejoy-patch-2` so you can test end-to-end and also changed the base branch for this PR to `jlovejoy-patch-1`. Now we just need to merge this PR first and then #387 .

src/app/views.py
```diff
@@ -117,7 +117,6 @@ def submitNewLicense(request):
                     licenseHeader = form.cleaned_data['licenseHeader']
                     licenseComments = form.cleaned_data['comments']
                     licenseText = form.cleaned_data['text']
-                    userEmail = form.cleaned_data['userEmail']
```
rtgdk, line 120: You need to remove all other instances of this variable `userEmail` used in the below code.
Like line no 149
```
data['userEmail'] = userEmail
```
then line 163
```
licenseRequest = LicenseRequest(licenseAuthorName=licenseAuthorName, fullname=licenseName, shortIdentifier=licenseIdentifier,
                        submissionDatetime=now, userEmail=userEmail, notes=licenseNotes, xml=xml)
```

rtgdk: @jlovejoy You can make `userEmail=""` (empty) or update the `LicenseRequest` model on 163 to pass userEmail as None/Empty. We can also change the underlying structure of `LicenseRequest` but let's not do it right now.

rtgdk (changes requested):

rtgdk:
@jlovejoy Requested some changes. then we can merge this on #387  and test everything together.

src/app/views.py
jlovejoy: see latest changes...

goneall:
@jlovejoy @rtgdk - Just checking on the status of this PR - going to do a release over the next few days

jlovejoy:
I think this is ready (at least from my somewhat naive perspective) @goneall 

thanks for the help @rtgdk 

I do still need to do more work (or some help) on the other one tho, but might be good to merge this first?

rtgdk:
@jlovejoy @goneall LGTM.
Merging this in `master` instead of `jlovejoy-patch-1` #387

jlovejoy:
Thanks @rtgdk ! #387 is a separate set of changes, so I'll try to get back to working on that now!

PR #391, opened 2022-09-09
Added the NTIA Conformance Checker to the SPDX Online Tools

linynjosh:
Signed-off-by: Josh Lin <linynjosh@gmail.com>

PR #392, opened 2022-09-26
Make /api/check_license more machine readable

autoantwort:
Currently all the information is inside a single string. 

The new output looks like the following:
```json
{"matched_license": null, "match_type": "No match", "all_matches": {}}
```
```json
{"matched_license": "Apache-2.0", "match_type": "Standard License match", "all_matches": {"Apache-2.0": 0.9813657538916032, "SHL-0.51": 0.9398223577022127, "SHL-0.5": 0.9392793360647127, "ECL-2.0": 0.958951030238159}}
```
```json
{"matched_license": "BSL-1.0", "match_type": "Perfect match", "all_matches": {"BSL-1.0": 1.0}}
```
Is there a reason while all the uploaded files are saved on disk and in the database? This seems very unnecessary and only consumes a lot of disk space over time.

### DIFF
--- a/src/app/utils.py
+++ b/src/app/utils.py
@@ -518,26 +518,22 @@ def check_spdx_license(licenseText):
     spdxLicenseTexts = r.mget(spdxLicenseIds)
     licenseData = dict(list(zip(spdxLicenseIds, spdxLicenseTexts)))
     matches = get_close_matches(licenseText, licenseData)
-
     if not matches:
         matchedLicenseIds = None
         matchType = 'No match'
-    
-    elif 1.0 in list(matches.values()) or all(0.99 < score for score in list(matches.values())):
-        matchedLicenseIds = list(matches.keys())
-        matchType = 'Perfect match'
-    
     else:
-        for licenseID in matches:
-            listedLicense = getListedLicense(licenseID)
-            isTextStandard = checkTextStandardLicense(listedLicense, licenseText)
-            if not isTextStandard:
-                matchedLicenseIds = licenseID
-                matchType = 'Standard License match'
-            else:
-                matchedLicenseIds = max(matches, key=matches.get)
-                matchType = 'Close match'
-    return matchedLicenseIds, matchType
+        matchedLicenseIds = max(matches, key=matches.get)
+        if matches[matchedLicenseIds] > 0.99:
+            matchType = 'Perfect match'
+        else:
+            matchType = 'Close match'
+            for licenseID in matches:
+                listedLicense = getListedLicense(licenseID)
+                isTextDifferent = checkTextStandardLicense(listedLicense, licenseText)
+                if not isTextDifferent:
+                    matchedLicenseIds = licenseID
+                    matchType = 'Standard License match'
+    return matchedLicenseIds, matchType, matches
 
 
 def getFileFormat(to_format):
--- a/src/app/utils.py
+++ b/src/app/utils.py
@@ -523,7 +523,7 @@ def check_spdx_license(licenseText):
         matchType = 'No match'
     else:
         matchedLicenseIds = max(matches, key=matches.get)
-        if matches[matchedLicenseIds] > 0.99:
+        if matches[matchedLicenseIds] == 1.0:
             matchType = 'Perfect match'
         else:
             matchType = 'Close match'
--- a/src/app/views.py
+++ b/src/app/views.py
@@ -127,7 +127,7 @@ def submitNewLicense(request):
                         # This is present only when executing submit license via tests
                         urlType = request.POST["urlType"]
 
-                    matchingIds, matchingType = utils.check_spdx_license(licenseText)
+                    matchingIds, matchingType, _ = utils.check_spdx_license(licenseText)
                     matches = ['Perfect match', 'Standard License match', 'Close match']
                     if matchingType in matches:
                         data['matchType'] = matchingType
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from django.http import JsonResponse
 from rest_framework.parsers import FileUploadParser,FormParser, MultiPartParser
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
 from api.models import ValidateFileUpload,ConvertFileUpload,CompareFileUpload,CheckLicenseFileUpload,SubmitLicenseModel
-from api.serializers import ValidateSerializer,ConvertSerializer,CompareSerializer,CheckLicenseSerializer,SubmitLicenseSerializer,ValidateSerializerReturn,ConvertSerializerReturn,CompareSerializerReturn,CheckLicenseSerializerReturn,SubmitLicenseSerializerReturn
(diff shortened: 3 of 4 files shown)


src/app/utils.py
```diff
@@ -518,26 +518,22 @@ def check_spdx_license(licenseText):
...
     else:
-        for licenseID in matches:
-            listedLicense = getListedLicense(licenseID)
-            isTextStandard = checkTextStandardLicense(listedLicense, licenseText)
-            if not isTextStandard:
-                matchedLicenseIds = licenseID
-                matchType = 'Standard License match'
-            else:
-                matchedLicenseIds = max(matches, key=matches.get)
-                matchType = 'Close match'
```
autoantwort, line 539: `Standard License match` were overwritten by `Close match` matches. But I assume that Standard License matches are better then close matches

src/app/utils.py
```diff
@@ -518,26 +518,22 @@ def check_spdx_license(licenseText):
...
     licenseData = dict(list(zip(spdxLicenseIds, spdxLicenseTexts)))
     matches = get_close_matches(licenseText, licenseData)
-
     if not matches:
         matchedLicenseIds = None
         matchType = 'No match'
-    
-    elif 1.0 in list(matches.values()) or all(0.99 < score for score in list(matches.values())):
-        matchedLicenseIds = list(matches.keys())
-        matchType = 'Perfect match'
```
autoantwort, line 528: This seems strange. If there is one match with a `1.0` we include all matches in the result (also bad matched). And if all matches are over `0.99` we return them all. In general, sometimes the function returns a list and sometimes single values.

goneall:
@rtgdk - Could you do a review on this PR over the next few days?  I'm updating the Java library over the next couple of days and if this PR looks good, I can include it in the release.

src/app/views.py
```diff
@@ -127,7 +127,7 @@ def submitNewLicense(request):
                         # This is present only when executing submit license via tests
                         urlType = request.POST["urlType"]
 
-                    matchingIds, matchingType = utils.check_spdx_license(licenseText)
+                    matchingIds, matchingType, _ = utils.check_spdx_license(licenseText)
```
rtgdk, line 130: Should we show all matches with percent match on UI as well? Maybe let's open a separate task for that.

src/app/utils.py
```diff
@@ -518,26 +518,22 @@ def check_spdx_license(licenseText):
...
-            else:
-                matchedLicenseIds = max(matches, key=matches.get)
-                matchType = 'Close match'
-    return matchedLicenseIds, matchType
+        matchedLicenseIds = max(matches, key=matches.get)
+        if matches[matchedLicenseIds] > 0.99:
+            matchType = 'Perfect match'
+        else:
+            matchType = 'Close match'
+            for licenseID in matches:
```
rtgdk, line 530: We are looping over all matched licenses list but returning only the last standard license in list or the one which close match. @goneall Should we return all the matched licenses or the license with the max match percent? We need some way to show match percentages vs license and let users choose?

src/app/utils.py
rtgdk: Yes, agreed. This is strange and incorrect.

rtgdk (commented):
@autoantwort Overall looks good to me. I am just confused about what to show in case of non-Perfect match. Should we show max matched license or all the matched licenses or the single standard license in a list of multiple non-standard license matches.
@goneall Please suggest

rtgdk:
@autoantwort No, files are not needed after the processing is done. We are saving the uploaded files to be used by the spdk java tools application. If there's a way we can instead send the file as a stream to the java application then we can use that. Or we can add cleanup check after the processing is done. Or a separate cleanup cron job.

autoantwort:
> @autoantwort No, files are not needed after the processing is done. We are saving the uploaded files to be used by the spdk java tools application. If there's a way we can instead send the file as a stream to the java application then we can use that. 

The java application already gets the license text as string. So what we currently do:
String in body -> write to file -> create database entry -> read file to String -> pass String to java tool
What we could do:
String in body -> pass String to java tool 

I have no idea why the first version was chosen :D

src/app/views.py
autoantwort: Probably, but yes, this is another task

src/app/utils.py
goneall: We could probably improve this.  Considerations would be the score and also if the SPDX listed license is deprecated - there are a lot of duplicate license texts where the license is deprecated.  If we match a deprecated and a non-deprecated license, we should only return the non-deprecated.

We may want to open a separate PR and/or Issue since this would be a pretty sizeable change.

src/app/utils.py
goneall: Correct

src/app/views.py
goneall: Agree as well - looks like some good improvements in this PR

goneall (approved):

goneall (changes requested):
I just tested out this code and it reported a perfect match for something that should be a close match.

In looking through the changes, it isn't obvious where the issue is.

I just removed one of the words `DISLAIMER` from the BSD license text.  In the current deployed version, it shows a "close match".  In this branch it shows a perfect match which is incorrect.

Below is the text that I used:
```
BSD Zero Clause License

Copyright (C) YEAR by AUTHOR EMAIL

Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```

autoantwort:
I guess `1.0 in list(matches.values()) or all(0.99 < score for score in list(matches.values()))` changed to 
```python
        matchedLicenseIds = max(matches, key=matches.get)
        if matches[matchedLicenseIds] > 0.99:
            matchType = 'Perfect match'
```
makes the difference if the list is `[0.99, 0.8]`. I have not tested it but I guess it is that. Maybe I should change it to `if matches[matchedLicenseIds] == 1.0:`

goneall:
> Maybe I should change it to if matches[matchedLicenseIds] == 1.0:

If you add a commit for this change, I can test it out (I have the environment all setup for testing).

autoantwort:
Done

goneall:
@autoantwort That works - thanks!

@rtgdk If you're OK with these changes, feel free to merge.

goneall (approved):

goneall:
@autoantwort It looks like PR #391 created a merge conflict with core.py in this PR.

From looking at the conflict, it looks like it is a bit more involved than just merging the 2 PR's - the design of the `check_license` function has changed.

Can you take a look at it and resolve the conflict?

goneall:
@autoantwort Never mind the merge conflict - I'm putting back the original code which should resolve (see PR #409).  I'll test this out one more time and merge.

PR #397, opened 2022-10-14
Update references to main

jlovejoy:
change master to main references to license list repo

@goneall - please check if this looks right!

### DIFF
--- a/src/app/utils.py
+++ b/src/app/utils.py
@@ -101,8 +101,8 @@ def makePullRequest(username, token, branchName, updateUpstream, fileName, commi
             }
     else:
         if(updateUpstream=="true"):
-            """ If user wants to update the forked repo with upstream master """
-            update_url = "{0}/git/refs/heads/master".format(TYPE_TO_URL_NAMESPACE[NORMAL] if is_ns else TYPE_TO_URL_LICENSE[NORMAL])
+            """ If user wants to update the forked repo with upstream main """
+            update_url = "{0}/git/refs/heads/main".format(TYPE_TO_URL_NAMESPACE[NORMAL] if is_ns else TYPE_TO_URL_LICENSE[NORMAL])
             response = requests.get(update_url, headers=headers)
             data = json.loads(response.text)
             sha = data["object"]["sha"]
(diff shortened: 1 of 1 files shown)


goneall:
@jlovejoy Thanks for the PR

In looking at the code - it looks like the same code is used for updates to the namespace repo.

Rather than fix this in the code, I think I'll rename the master branch to main in the namespace repo.  @jlovejoy @swinslow Let me know if you have any concerns with this approach.

jlovejoy:
sounds fine to me, @goneall 

is that why this PR is failing?

goneall:
> is that why this PR is failing?

No - there are some known issues with the unit tests when running under the CI.  I end up having to look at each test result to see if a new error is introduced - rather inconvenient.  I noted the above just by reviewing the code.

goneall:
I found 3 repos where we need to rename the default branch to match the license-list-XML repo.  I created issues for each of these to track - I'll wait a week or so just to make sure no one has any issues with the renaming - then I'll fix:

- https://github.com/spdx/license-namespace-test/issues/9
- https://github.com/spdx/TEST-LicenseList-XML/issues/122
- https://github.com/spdx/license-namespace/issues/5

swinslow:
@goneall No concerns from me with the branch renaming, sounds good.

rtgdk:
@goneall If we are on track for renaming `master` to `main`, we should also consider https://github.com/spdx/spdx-online-tools 😛 . Opened an issue for future #401

goneall:
I've renamed the default branches for all the repos referenced in the online tools.

@jlovejoy - could you change the other references to `master` to refer to `main`?

Once that's done, I can give it a test and see if it solves the issue.

goneall:
> If we are on track for renaming master to main, we should also consider https://github.com/spdx/spdx-online-tools 😛 . Opened an issue for future https://github.com/spdx/spdx-online-tools/issues/401

@rtgdk Good point - I should do the same for the Java repos.

jlovejoy:
@goneall - updated 3 more references, I think that's all of them now. This included a link to the license-list-data repo as well.

goneall:
@jlovejoy I rebased this PR with the fix for the docker problem - you should be able to just pull down the latest and test it out.

PR #400, opened 2022-10-18
Remove the namespace menu items

goneall:
Removing the namespace feature for now since the namespace functionality has not be decided.

Once we decide on the namespace functionality, we can re-enable the menu.


rtgdk (approved):
LGTM, we are only disabling UI usage, not the API.

PR #410, opened 2022-11-29
Remove CheckLicenseFileUpload model

autoantwort:
Follow up of #392.

The model is not needed to get a license identifier from a text file.


src/api/migrations/0002_auto_20221129_1430.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ('api', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submitlicensemodel',
+            name='osiApproved',
+            field=models.CharField(choices=[(0, '-'), ('Approved', 'Approved'), ('Not Submitted', 'Not Submitted'), ('Pending', 'Submitted, but pending'), ('Rejected', 'Rejected'), ('Unknown', "Don't know")], default=0, max_length=15),
+        ),
```
autoantwort, line 17: It seems that someone changes the model but have not created a migration for that.

src/api/models.py
```diff
@@ -61,13 +61,6 @@ class CompareFileUpload(models.Model):
     rfilename = models.CharField(max_length=32,null=False,blank=False)
     status = models.IntegerField(default=200,blank=False)
 
-class CheckLicenseFileUpload(models.Model):
```
rtgdk, line 64: This is being used in some of the tests as well. You need to remove the usage for this from there as well (tests.py)

goneall: I added issue #422 to track fixing the unit tests.  Since the unit tests are already failing for other reasons, I'll go ahead and merge this PR and hopefully fix the unit tests in an upcoming release.

PR #419, opened 2023-01-08
Update NTIA conformance checker to version 0.1.0

goneall:
Also adds the NTIA conformance checker version to the about page

### DIFF
--- a/src/src/version.py
+++ b/src/src/version.py
@@ -9,3 +9,4 @@ spdx_online_tools_version = '1.0.9'
 Visit https://github.com/spdx/tools-java/releases to know about the tools releases.
 """
 java_tools_version = '1.1.3'
+ntia_conformance_checker_version = '0.1.0'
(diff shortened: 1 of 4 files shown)


src/src/version.py
```diff
@@ -9,3 +9,4 @@
 Visit https://github.com/spdx/tools-java/releases to know about the tools releases.
 """
 java_tools_version = '1.1.3'
+ntia_conformance_checker_version = '0.1.0'
```
rtgdk, line 12: I believe there should be a way to get this version from `ntia-conformance-checker` package itself when it is released. Something like `from ntia-conformance-checker import version` or `ntia_conformance_checker.version`. We can dig into this later on. Similarly for others.

rtgdk (approved):
LGTM

src/src/version.py
goneall: Really good point - I'll create a separate issue to add this in later

Issue #344, opened 2022-03-26
Python3 Unit Test Failure  test_submitlicense_api_with_invalid_code


goneall:
The following unit test error occured on the Python 3 branch:

```
======================================================================
ERROR: test_submitlicense_api_with_invalid_code (api.tests.SubmitLicenseModelsTests)
Post submit license api with empty authentication code
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\gary\git\spdx-online-tools\src\api\tests.py", line 403, in test_submitlicense_api_with_invalid_code
    self.assertIn("No authentication code provided.",resp3.content)
  File "c:\Python38\lib\unittest\case.py", line 1176, in assertIn
    if member not in container:
TypeError: a bytes-like object is required, not 'str'
```

Issue #343, opened 2022-03-26
Python3 Unit Test failure test_submitlicense_api_with_invalid_fields


goneall:
The following unit test error occured on the Python 3 branch:

```
======================================================================
ERROR: test_submitlicense_api_with_invalid_fields (api.tests.SubmitLicenseModelsTests)
Test for invalid serializer
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\gary\git\spdx-online-tools\src\api\tests.py", line 440, in test_submitlicense_api_with_invalid_fields
    self.assertIn("is not a valid choice.",resp9.content)
  File "c:\Python38\lib\unittest\case.py", line 1176, in assertIn
    if member not in container:
TypeError: a bytes-like object is required, not 'str'
```

Issue #341, opened 2022-03-26
Python3 Unit Test Failure test_generate_xml


goneall:
Below is the error reported by the unit test:

```
======================================================================
FAIL: test_generate_xml (api.tests.SubmitLicenseModelsTests)
View for generating an xml from license submittal form fields
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\gary\git\spdx-online-tools\src\api\tests.py", line 398, in test_generate_xml
    self.assertEqual(self.xml, xml)
AssertionError: '<SPD[50 chars]nse"> <license isOsiApproved="false" licenseId[1038 chars]on> ' != '<SPD[50 chars]nse"><license isOsiApproved="false" licenseId=[1024 chars]ion>'
Diff is 4529 characters long. Set self.maxDiff to None to see it.
```

PR #438, opened 2023-02-20
Fix SubmitLicenseModelsTests tests in python3 branch

HirumalPriyashan:
Addressed followings,

- Fix https://github.com/spdx/spdx-online-tools/issues/341
- Fix https://github.com/spdx/spdx-online-tools/issues/343
- Fix https://github.com/spdx/spdx-online-tools/issues/344

### DIFF
--- a/src/api/tests.py
+++ b/src/api/tests.py
@@ -353,7 +353,7 @@ class SubmitLicenseModelsTests(APITestCase):
         self.userEmail = "test@mail.com"
         self.licenseAuthorName = ""
         self.listVersionAdded = ""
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - Thanks @HirumalPriyashan

HirumalPriyashan:
Hi @goneall, after this PR got merged I have started on refactoring the `tests.py` in `api` app to reduce the file size.
But following tests are continued fail in main branch. Is this also happening in remote environment as well, or is it something due to wrong config in my environment.

```
FAIL: test_compare_api (api.tests.CompareFileUploadTests)
Access get without login
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home//spdx-online-tools/src/api/tests.py", line 249, in test_compare_api
    self.assertEqual(resp3.status_code,200)
AssertionError: 400 != 200

======================================================================
FAIL: test_validate_api (api.tests.ValidateFileUploadTests)
Access get without login
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home//spdx-online-tools/src/api/tests.py", line 85, in test_validate_api
    self.assertEqual(resp4.status_code,200)
AssertionError: 400 != 200
```

Also there was a need to remove import and usage of `CheckLicenseFileUpload` since that available in the `models` anymore which was producing errors in imports.
https://github.com/spdx/spdx-online-tools/blob/400bbff988b5cb0abc9e3d7813229850b3ff9fd4/src/api/tests.py#L35

I can send the current progress of test refactoring as a draft PR if it is easier for review.

goneall:
@HirumalPriyashan I'm not completely sure on the unit test failures.  I recall seeing similar failures in the past, but it was a while ago and I don't recall the details.

@rtgdk - Can you take a look and provide some feedback?

rtgdk:
`test_compare_api` tests the post api - `api/compare` after logging in with the default users used in `setUp` of `CompareFileUploadTests`. Check whether your username/password is not changed and should work without issue.

Next the api calls the post method on `/api/compare`, check `def compare` in `api/views.py` and where exactly it failing. The post request on views. py doesn't seem to check for user login but it interacts with different models and then set the return status.

HirumalPriyashan:
@rtgdk Thanks for the explanation, I will look into that.

Issue #328, opened 2021-08-21
Add clean up disk space feature


goneall:
The application is taking an increasing amount of disk space on the production instance.

Most of the space is consumed in the /media folder - /media/AnonymousUser being the largest at 2GB.

Suggest we add some type of regular clean-up of the media folders.  This could be added into the app itself or a separate CRON job we could install via the Docker files.

rtgdk:
The folder stores the license files uploaded by the user (for validating/comparing/etc). Old files are generally not needed. When user are not logged in, the files are uploaded to `AnonymousUser`. I think cron job should delete more than 10 days old files. The api also creates some files which are returned to the user for downloading.

rtgdk:
@autoantwort No files are not needed after the processing is done. We are saving the uploaded files to be used by the spdk java tools application. If there's a way we can instead send the file as a stream to the java application then we can use that. Or we can add cleanup check after the processing is done. Or a separate cleanup cron job

BassCoder2808:
Hi @goneall I would be happy to contribute to this issue if it is still open.
Going by the approach which @rtgdk mentioned above, I have made a small CRON job for the same, let me know if it is correct or needs any improvement if this issue is till open
```
import os
import time

MEDIA_DIR = '/media/AnonymousUser'
DAYS_THRESHOLD = 10

now = time.time()

for filename in os.listdir(MEDIA_DIR):
    filepath = os.path.join(MEDIA_DIR, filename)
    if os.path.isfile(filepath) and (now - os.stat(filepath).st_mtime) > (DAYS_THRESHOLD * 86400):
        os.remove(filepath)

```

goneall:
@BassCoder2808 - Thanks!  The above approach looks good to me.

We could add the script in the `scripts` folder and update the [dockerfile.prod](https://github.com/spdx/spdx-online-tools/blob/main/dockerfile.prod) file to install and run the script.

Not quite sure the best way to test the changes before pushing to production - let me know if you have any ideas.

Please feel free to create a PR.

BassCoder2808:
Sure @goneall I will create a pr for the same, and will also try how we can check if the code works correctly or not

BassCoder2808:
@goneall I have written test case for the same, let me know your thoughts on the same, I have not added this in the above pr
```
import datetime
import os
from django.test import TestCase

class TestCronJob(TestCase):
    def test_delete_old_files(self):
        # create a test directory with some files
[...]

PR #440, opened 2023-03-10
Add clean up disk space feature

BassCoder2808:
Intended to solve #328 
1. Inside the `scripts` folder a new cleanup.py file has been added
2. In the base `spdx` folder `crontab` file has been added which will be used to run our corn job
3. `dockerfile.prod` has been changed to add the corn job and run the python script


rtgdk:
This looks a good way to do it. But there are django specific model cleanups/media cleanup libraries (`django-cleanup` `django-unused-media` ) which we can use. Let's try searching for a much simpler solution inside django itself which is the creator for these media files.

Also, some models keep a reference to these files so maybe deleting them as well is needed or archiving them.

So let's try to find a common solution for both django models (deleting/archiving/removing file reference if any) and deleting the media files asynchronously.

rtgdk:
This looks a good way to do it. But there are django specific model cleanups/media cleanup libraries (`django-cleanup` `django-unused-media` ) which we can use. Let's try searching for a much simpler solution inside django itself which is the creator for these media files.

Also, some models keep a reference to these files so maybe deleting them as well is needed or archiving them.

So let's try to find a common solution for both django models (deleting/archiving/removing file reference if any) and deleting the media files asynchronously.

rtgdk:
If the django based solution is not required then this cron job should do the work. Thanks for the contribution @BassCoder2808

BassCoder2808:
Sure @rtgdk I will try searching for some solutions inside Django itself which can help, but the issue actually mentioned a corn job so went with the above solution

rtgdk:
Yeah that was my initial idea but new libs came up inside django itself. In worst case, cron job seems the way to go.

src/app/tests.py
```diff
@@ -1543,3 +1544,28 @@ def test_no_license_id_on_license_requests_edit_xml(self):
...
+            with open(file_path, 'w') as f:
+                f.write('test')
+            # set file creation time to 11 days ago
+            creation_time = datetime.datetime.now() - datetime.timedelta(days=11)
+            if i<=5:
+                os.utime(file_path, (creation_time.timestamp(), creation_time.timestamp()))
+
+        # run the cron job to delete old files
+        # (you may need to modify this depending on how you're running the cron job)
+        os.system('python manage.py delete_old_files')
```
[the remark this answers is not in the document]
BassCoder2808: @sonatype-lift ignore

BassCoder2808:
> This looks a good way to do it. But there are django specific model cleanups/media cleanup libraries (`django-cleanup` `django-unused-media` ) which we can use. Let's try searching for a much simpler solution inside django itself which is the creator for these media files.

Hi @rtgdk I checked the libraries which you had mentioned, I think if we want we can make use of `django-cleanup` rather than `django-unused-media` as it is regularly being updated as compared to the other. But one problem which I figured out was that we will need to change the model for file field. Also will need some help in understanding the system, as this library has some limitations which they have mentioned the documentation [django-cleanup](https://github.com/un1t/django-cleanup).

I fell that currently, we can go forward with the corn job, and then once we have setup this, we can change and remove the cron job in the future

src/app/management/commands/delete_old_files.py
```diff
@@ -0,0 +1,27 @@
+import datetime
```
rtgdk, line 1: Do we need this file now?

scripts/cleanup.py
```diff
@@ -0,0 +1,12 @@
+import os
+import time
+
+MEDIA_DIR = './src/app/media/AnonymousUser'
+DAYS_THRESHOLD = 10
+
+now = time.time()
+
+for filename in os.listdir(MEDIA_DIR):
```
rtgdk, line 9: Let's add a logger to track time of cron job and also log the files it's deleting and date of the file. Also, cron should log to that file by default.

src/app/tests.py
```diff
@@ -1543,3 +1544,27 @@ def test_no_license_id_on_license_requests_edit_xml(self):
         self.assertEqual(resp.status_code,200)
         self.assertIn("app/license_namespace_requests.html",(i.name for i in resp.templates))
         self.assertEqual(resp.resolver_match.func.__name__,"licenseNamespaceRequests")
+
+class TestCronJob(TestCase):
+    def test_delete_old_files(self):
```
rtgdk, line 1549: We can directly run test on cleanup.py script

rtgdk (changes requested):

scripts/cleanup.py
BassCoder2808: Ok, I will add that and update the PR

src/app/management/commands/delete_old_files.py
BassCoder2808: Yes, actually for the testing I have made a django-admin command for that we are requiring this

src/app/tests.py
BassCoder2808: Ok, I will change the unit test as well, actually currently it was being run through a django-admin command, I will change that

BassCoder2808: I have changed the tests now, they now directly use the cleanup.py script instead of the django commands
image

scripts/cleanup.py
```diff
@@ -0,0 +1,13 @@
+import os
```
rtgdk, line 1: Let's move this to app folder as this scripts works only on app media directory for now.

src/app/tests.py
```diff
@@ -27,6 +28,11 @@
 from social_django.models import UserSocialAuth
 from django.conf import settings
 import os
+import sys
+# assuming the Django project directory is at the same level as the scripts directory
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', 'scripts')))
```
rtgdk, line 33: I think with `scripts` inside `app`, we won't need this?

src/app/tests.py
```diff
@@ -1543,3 +1549,25 @@ def test_no_license_id_on_license_requests_edit_xml(self):
         self.assertEqual(resp.status_code,200)
         self.assertIn("app/license_namespace_requests.html",(i.name for i in resp.templates))
         self.assertEqual(resp.resolver_match.func.__name__,"licenseNamespaceRequests")
+
+class TestCronJob(TestCase):
+    def test_delete_old_files(self):
+        """Check if the files older than 10 days are getting deleted or not"""
+        # create a test directory with some files
+        test_dir = os.path.join(os.getcwd() + '/app/media', 'AnonymousUser')
```
rtgdk, line 1557: You can use Media folder constant from settings.py I guess

src/app/tests.py
```diff
@@ -1543,3 +1549,25 @@ def test_no_license_id_on_license_requests_edit_xml(self):
...
+        os.makedirs(test_dir, exist_ok=True)
+        for i in range(1, 11):
+            file_path = os.path.join(test_dir, f'test_file_{i}.txt')
+            with open(file_path, 'w') as f:
+                f.write('test')
+            # set file creation time to 11 days ago
+            creation_time = datetime.datetime.now() - datetime.timedelta(days=11)
```
rtgdk, line 1564: Let's keep some files newer than 10 days as well and assert for them.

src/app/tests.py
```diff
@@ -1543,3 +1549,25 @@ def test_no_license_id_on_license_requests_edit_xml(self):
...
         self.assertEqual(resp.resolver_match.func.__name__,"licenseNamespaceRequests")
+
+class TestCronJob(TestCase):
+    def test_delete_old_files(self):
+        """Check if the files older than 10 days are getting deleted or not"""
+        # create a test directory with some files
+        test_dir = os.path.join(os.getcwd() + '/app/media', 'AnonymousUser')
+        os.makedirs(test_dir, exist_ok=True)
+        for i in range(1, 11):
+            file_path = os.path.join(test_dir, f'test_file_{i}.txt')
```
rtgdk, line 1560: Better to store this in array and directly compare it they exists or not. Creating `f'test_file_{i}.txt'` filename like this should be done at one place so that maintenance is easy and can changed quickly.

rtgdk (commented):
@BassCoder2808 Overall LGTM, left some comments to address before merge.

scripts/cleanup.py
BassCoder2808: Ok, I will update the PR for the same and will let you know once done

src/app/tests.py
BassCoder2808: Now I have updated the unit tests, like 5 files are older than 10 days and others are not and they test is passing as well

image

src/app/tests.py
BassCoder2808: Like this is for the test files only, we are first creating the 10 test files and then based on whether they were deleted or not, we are asserting, so should I direct assert with an array at the end instead of the for loop?

rtgdk (approved):
LGTM

Issue #308, opened 2021-05-07
Error when submitting a license.


ghost:
Error when submitting a license, now all my progress is gone. oh well.

`Unexpected error, please email the SPDX technical workgroup that the following error has occurred: Traceback (most recent call last): File "/spdx/src/app/views.py", line 168, in submitNewLicense licenseRequest.save() File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 808, in save force_update=force_update, update_fields=update_fields) File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 838, in save_base updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields) File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 924, in _save_table result = self._do_insert(cls._base_manager, using, fields, update_pk, raw) File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 963, in _do_insert using=using, raw=raw) File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 85, in manager_method return getattr(self.get_queryset(), name)(*args, **kwargs) File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 1079, in _insert return query.get_compiler(using=using).execute_sql(return_id) File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 1112, in execute_sql cursor.execute(sql, params) File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute return self.cursor.execute(sql, params) File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__ six.reraise(dj_exc_type, dj_exc_value, traceback) File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute return self.cursor.execute(sql, params) DataError: value too long for type character varying(35)`

jlovejoy:
@goneall - has this been addressed otherwise?

goneall:
@jlovejoy As far as I know, this is still an issue.  We should validate all field lengths.  Should be a relatively easy task for someone familiar with Django (which I'm not).

BassCoder2808:
Hi @goneall I would like to contribute to this issue if it is still open, can you guide me through like what needs to be done to solve this, it would help me a lot

goneall:
[...]

PR #443, opened 2023-03-11
Error when submitting a license

BassCoder2808:
Solves #308 
Added validation checks in the `forms.py` for the `userEmail` field


dockerfile.prod
```diff
@@ -58,4 +58,4 @@ EXPOSE 8000
 RUN mv ./supervisor_api.conf /etc/supervisor/conf.d
 
 # Start the supervisor
-CMD /usr/bin/supervisord -n -c /etc/supervisor/supervisord.conf
+CMD /usr/bin/supervisord -n -c /etc/supervisor/supervisord.conf
```
goneall, line 61: It looks like the last linefeed was removed from the dockerfile.prod file - if you could put this back it would make for a cleaner pull request

goneall (changes requested):
One minor fix requested

goneall (approved):
Thanks @BassCoder2808

Issue #202, opened 2020-07-15
Unit test failure - test_post_submit


goneall:
The following test_post_submit unit test fails on the master branch:

```
======================================================================
FAIL: test_post_submit (app.tests.SubmitNewLicenseViewsTestCase)
POST Request for submit a new license
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/spdx-online-tools/spdx-online-tools/src/app/tests.py", line 1229, in test_post_submit
    self.assertEqual(resp.status_code,200)
AssertionError: 500 != 200
```

Ugtan:
@goneall I'm getting a different error in my machine.

![image](https://user-images.githubusercontent.com/32966387/87609880-63c62680-c721-11ea-9147-b92f125c005e.png)

FYI, All the test cases that are failing are the once that were skipped because they require github credentials which is why these tests were not tested. I will take a look at it.

goneall:
> FYI, All the test cases that are failing are the once that were skipped because they require github credentials which is why these tests were not tested. I will take a look at it.

I added the credentials to Git a couple weeks back - this may have exposed an issue that was introduced some time ago.  The errors are coming from the automated Git actions on checkin.

@Ugtan Thanks for looking into the issue.

SidharajYadav:
How i can contribute
please guide me
I like to solve this problem

BanulaKumarage:
Hi @goneall can you clarify the difference between test_post_submit and test_submit_new_license

goneall:
@BanulaKumarage Probably a better question for @rtgdk - I'm not as familiar with the unit test code

BanulaKumarage:
Hi @goneall I made a pr to fix this issue. Can you review it?

rtgdk:
@BanulaKumarage `test_submit_new_license` is the test for get request on submit-new-license page while `test_post_submit` tests the post request. Both of them have different functionalities
- get -> retrivies page and check the form is displayed with correct values
- post -> checks with correct data license is submitted successfully

BanulaKumarage:
Thank you @rtgdk. I got the idea

jlovejoy:
@rtgdk - what is the status of this?

bact:
This test is currently skipped when there is not GitHub credentials provided.

https://github.com/spdx/spdx-online-tools/blob/a38a0534f7534679906943566e811c316216053d/src/app/tests.py#L1343-L1345

bact:
[...]

PR #444, opened 2023-03-11
#202 Fixing Unit test failure - test_post_submit

BanulaKumarage:
- Two required fields (exampleUrl,  comments) were missing in the request and they were added.
- Byte type patterns were not converted to utf-8 strings in line 518 and 519 in [src/app/utils.py](https://github.com/spdx/spdx-online-tools/blob/main/src/app/utils.py) (This error was raised when the first point is fixed)
image


rtgdk:
LGTM.

Issue #200, opened 2020-07-15
Unit test failure


goneall:
The following unit test is failing on the current master branch:

```
======================================================================
ERROR: test_archive_license_requests_feature (app.tests.ArchiveLicenseRequestsViewsTestCase)
Check if the license is shifted to archive requests when archive button is pressed
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/spdx-online-tools/spdx-online-tools/src/app/tests.py", line 1123, in test_archive_license_requests_feature
    if driver.find_element_by_id('archive_button' + str(license_obj.id)):
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 360, in find_element_by_id
    return self.find_element(by=By.ID, value=id_)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 978, in find_element
    'value': value})['value']
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
NoSuchElementException: Message: Unable to locate element: [id="archive_button1"]
```

rtgdk:
This was introduced after https://github.com/spdx/spdx-online-tools/pull/154 got merged. This has already been raised at https://github.com/spdx/spdx-online-tools/issues/177 @anshuldutt21 Would you please fix this? Looks like if..else won't work. You would have to use try..catch. Also, please ensure that this test runs for admin user.

nishantc7:
@rtgdk Could you please help me out, I was trying to work on this but can't seem to work it through. I tried logging in first using socialauth, but the webdriver still can't locate the archive button. It should work once logged in right?

rtgdk:
[...]

PR #445, opened 2023-03-11
#200 Fixing Unit test failure - test_archive_license_requests_feature

BanulaKumarage:
- Changed if-else block to try a try-catch.
image

### DIFF
--- a/src/app/tests.py
+++ b/src/app/tests.py
@@ -1146,10 +1146,11 @@ class SubmitNewLicenseViewsTestCase(TestCase):
     def initialise(self):
         self.fullname = "BSD Zero Clause License"
         self.shortIdentifier = "0BSD"
+        self.exampleUrl = "testUrl"
         self.sourceUrl = "http://landley.net/toybox/license.html"
         self.urls = [self.sourceUrl]
         self.osiApproved = "no"
-        self.comments = ""
+        self.comments = "Test Comment"
         self.notes = ""
         self.licenseHeader = ""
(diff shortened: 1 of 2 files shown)


src/app/tests.py
```diff
@@ -1115,11 +1116,11 @@ def test_archive_license_requests_feature(self):
...
         self.assertEqual(license_name, "BSD Zero Clause License-00")
         self.assertEqual(LicenseRequest.objects.get(id=license_obj.id).archive, False)
-        if driver.find_element_by_id('archive_button' + str(license_obj.id)):
+        try:
             driver.find_element_by_id('archive_button' + str(license_obj.id)).click()
             driver.find_element_by_id('confirm_archive').click()
             self.assertEqual(LicenseRequest.objects.get(id=license_obj.id).archive, True)
-        else:
+        except NoSuchElementException:
             pass
```
rtgdk, line 1124: If the archive license request element doesn't exist then also the test will pass. This is not the correct fix. We should check why the `archive_button` is not showing up in the test and fix acccordingly.

BanulaKumarage: I will work on it @rtgdk

BanulaKumarage: Hi @rtgdk. I the archive/ unarchive button is visible only when the user is logged in with GitHub according to the following lines in src\app\templates\app\license_requests.html. So can we check whether the user is logged in with GitHub in this test?
```
{% if github_login and authorized %}
              <button id='archive_button{{elem.id}}' class='btn btn-success' width='25px' value='Archive'>Archive</button> {% endif %}
```

rtgdk: Yes, mocking github login would be the way to go @BassCoder2808

BanulaKumarage: I will try that way @rtgdk. btw @BanulaKumarage here (Incase any miscommunication with @BassCoder2808. No offence).

rtgdk: Apologies @BanulaKumarage . The autocompletion on `Ba` fooled me 😃

BanulaKumarage: No worries @rtgdk 😁

BanulaKumarage:
@rtgdk I made a new PR by mocking the authentication and fixing the tests for both archive and unarchive. Can you review PR #457?

rtgdk:
@BanulaKumarage I think we can close this PR

BanulaKumarage:
@rtgdk Closing the PR

Issue #449, opened 2023-03-22
Redirect on Successful License Submission


BassCoder2808:
Whenever we submit a new license, we get redirected to the same page after successful submission, instead, it should take the user to the issues page as discussed with @jlovejoy.

jlovejoy:
okay, I just tested this b/c I couldn't remember (and sometimes I don't notice these things because I just know where to go!) and yeah, it should not just stay on the submission page. I think going to the newly created GitHub issue would be best.

jlovejoy:
as discussed, have newly created Github issue open in new tab/window

BassCoder2808:
Yes, I had made changes to the PR as well #450, @rtgdk we can merge that if nothing else is needed in that

jlovejoy:
ok, we'll close issue once PR is merged

jlovejoy:
so, this was added, but the new window was blocked by my browser's pop-up blocker... maybe we should consider how common that might be?

BassCoder2808:
Hi @jlovejoy yes for the first time the pop-up window will be blocked for everyone by the browser itself, but once you allow it, it will not ask again. Actually, this is a security feature in every browser by default, so for the first time the user will need to allow pop-ups from `tools.spdx.com` to open

goneall:
@jlovejoy @BassCoder2808 - I'm going to close this as the original issue was resolved with PR #450 

If the pop-up blocker is an issue, please open a new issue.

PR #450, opened 2023-03-24
Fixing Redirect on Successful License Submission

BassCoder2808:
Intended to Solve #449

### DIFF
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -313,7 +313,7 @@ def submit_license(request):
             if 'urlType' in request.POST:
                 # This is present only when executing submit license via tests
                 urlType = request.POST["urlType"]
-            statusCode = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
+            statusCode, githubIssueId = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
             if str(statusCode) == '201':
                 result = "Success! The license request has been successfully submitted."
                 query.result = result
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -353,13 +353,14 @@ function scrollUpAndHighlight(element) {
               success: function (data) {
                 var githubCode = data.statusCode;
                 var githubIssueId = data.issueId;
+                var githubUrl = "https://github.com/spdx/licnse-list-xml/issues/";
                 if(githubCode == '201'){
                   var successMessage = "The license request has been successfully submitted.";
                   $("#messages").html('<div class="alert alert-success alert-dismissable fade in"><a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a><strong>Success! </strong>'+ successMessage +'</div>');
                   setTimeout(function() {
                     $("#messages").html("");
                   }, 7000);
-                  window.open("https://github.com/spdx/licnse-list-xml/issues/" + githubIssueId, "_blank");
+                  window.open(githubUrl + githubIssueId, "_blank");
                 }
                 else if(githubCode == '409'){
                   var matchType = data.matchType;
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -352,9 +352,11 @@ function scrollUpAndHighlight(element) {
               data: form,
               success: function (data) {
                 var githubCode = data.statusCode;
+                var githubIssueId = data.issueId;
                 if(githubCode == '201'){
                   var successMessage = "The license request has been successfully submitted.";
                   $("#messages").html('<div class="alert alert-success alert-dismissable fade in"><a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a><strong>Success! </strong>'+ successMessage +'</div>');
+                  window.location.href = "https://github.com/spdx/licnse-list-xml/issues/" + githubIssueId;
                       setTimeout(function() {
                         $("#messages").html("");
                       }, 7000);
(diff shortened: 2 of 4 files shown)


src/app/templates/app/submit_new_license.html
```diff
@@ -352,9 +352,11 @@
               data: form,
               success: function (data) {
                 var githubCode = data.statusCode;
+                var githubIssueId = data.issueId;
                 if(githubCode == '201'){
                   var successMessage = "The license request has been successfully submitted.";
                   $("#messages").html('<div class="alert alert-success alert-dismissable fade in"><a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a><strong>Success! </strong>'+ successMessage +'</div>');
+                  window.location.href = "https://github.com/spdx/licnse-list-xml/issues/" + githubIssueId;
```
rtgdk, line 359: I don't think directly transferring users to a new website is ideal. We should either redirect to our own license ID page which should list down the github url (i think it doesn't now) or open this github url in a new tab

BassCoder2808: Yes @rtgdk , I had also suggested the same idea like once there is a successful license submission the user should get redirected to 'https://tools.spdx.org/app/license_requests/{id}' wherein they can see more about his license instead of just again redirecting them to `submit new license`. [Discussion](https://github.com/spdx/license-list-XML/issues/1889) but @jlovejoy suggested that currently the working is like it gets redirected to the github page which was not working when I tried it.

rtgdk: Yeah opening github link looks fine - both technically and legally. But this should get opened in a new tab. The user should still have a tab to navigate spdx tools.

BassCoder2808: Ok, I will update the PR according so that whenever the new license get's submitted it gets opened in a new tab

src/app/views.py
```diff
@@ -162,7 +163,7 @@ def submitNewLicense(request):
                         licenseId = licenseRequest.id
                         serverUrl = request.build_absolute_uri('/')
                         licenseRequestUrl = os.path.join(serverUrl, reverse('license-requests')[1:], str(licenseId))
-                        statusCode = utils.createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseExamples, licenseRequestUrl, token, urlType)
+                        statusCode, githubIssueId = utils.createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseExamples, licenseRequestUrl, token, urlType)
```
rtgdk, line 166: I see multiple calls to `utils.createIssue`. You will need to update it elsewhere as well.

rtgdk (changes requested):

src/app/views.py
BassCoder2808: Hi @rtgdk , I have made the above changes, please let me know, if anything else needs to be changed in this

src/app/templates/app/submit_new_license.html
```diff
@@ -352,12 +352,14 @@
...
                 if(githubCode == '201'){
                   var successMessage = "The license request has been successfully submitted.";
                   $("#messages").html('<div class="alert alert-success alert-dismissable fade in"><a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a><strong>Success! </strong>'+ successMessage +'</div>');
-                      setTimeout(function() {
-                        $("#messages").html("");
-                      }, 7000);
+                  setTimeout(function() {
+                    $("#messages").html("");
+                  }, 7000);
+                  window.open("https://github.com/spdx/licnse-list-xml/issues/" + githubIssueId, "_blank");
```
rtgdk, line 362: nit: Let's have this as constant - `https://github.com/spdx/licnse-list-xml/issues/`

rtgdk (changes requested):
@BassCoder2808 I see one more use of createIssue in `api/views.py`. Please make changes, also left a minor comment. Will merge post these.

src/app/utils.py
```diff
@@ -352,7 +352,7 @@ def createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseCommen
     headers = {'Authorization': 'token ' + token}
     url = "{0}/issues".format(TYPE_TO_URL_LICENSE[urlType])
     r = requests.post(url, data=json.dumps(payload), headers=headers)
-    return r.status_code
+    return r.status_code, r.json()["number"]
```
rtgdk, line 355: Will r.json() always work without issues? What about when status_code is not as expected like 404, 400, will r.json() parse correctly. Also, will "number" be always in the json? If not let's parse status_code for successfull code list and parse `number` key if exists. @BassCoder2808

BassCoder2808: No actually, the number parameter will not be there if there was any issue while creating it through the API, I have added the code to handle that now, let me know if anything else needs to be done in that.

src/api/views.py
```diff
@@ -313,7 +313,7 @@ def submit_license(request):
             if 'urlType' in request.POST:
                 # This is present only when executing submit license via tests
                 urlType = request.POST["urlType"]
-            statusCode = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
+            statusCode, githubIssueId = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
```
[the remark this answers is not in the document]
BassCoder2808: @sonatype-lift ignoreall

goneall:
Looks like @rtgdk comment has been resolved - I'll go ahead and merge to include in the next update

src/app/templates/app/submit_new_license.html
```diff
@@ -352,12 +352,15 @@
               data: form,
               success: function (data) {
                 var githubCode = data.statusCode;
+                var githubIssueId = data.issueId;
+                var githubUrl = "https://github.com/spdx/licnse-list-xml/issues/";
```
whot, line 356: Note that there is a typo here resulting in a 404 on redirect: `licnse`

BassCoder2808: Thanks @whot for pointing out the error, I will crate a PR to solve this

PR #451, opened 2023-03-24
Adding Swagger Endpoint

BanulaKumarage:
@rtgdk I have added a swagger point and it contains all the API endpoints as the first step of #91. Need suggestions.


requirements.txt
```diff
@@ -2,6 +2,7 @@ Django==3.2.18
 jpype1==1.3.0
 numpy==1.22.0
 djangorestframework==3.12.2
+drf-yasg==1.21.5
```
rtgdk, line 5: I think we were not able to generate the required swagger specs using this last time in https://github.com/spdx/spdx-online-tools/pull/104. Can you please have a look at what all is generated right now and is in the correct shape. Post a screenshot of the swagger specs generated as well.

Or I would suggest to dig into other libraries which can do the same.
Just an idea - If we have swagger specs generated in some yaml format, we can host it in github pages/wiki/io website.

BanulaKumarage: Currently, the swagger endpoint can be accessed with all the API endpoints in the `api/swagger` endpoint. I'm working on generating the yaml. So far it's working with the endpoint. Will generate the yaml and update the PR. If the issue raises, I will try another library.

@rtgdk needs clarification. Why does the API documentation need to be hosted on GitHub pages? With this implementation, we can access the documentation with the endpoint.

rtgdk: @BanulaKumarage  Yeah hosting swagger specs directly where the site is hosted is fine.
I remember seeing this when we did the last implementation of drf-yasg, we saw some issues with the generation of documentation for fields in the api.
Can you check if every api has defined doc for every field? The actual documentation is currently hosted here - https://github.com/spdx/spdx-online-tools/wiki/REST-API-Fields-Request-and-Response - but is hard to find and also hard to track and update. Integrating directly into the website as a separate page looks perfect.

rtgdk: @BanulaKumarage Also, looks like some extra apis are generated as well. We should limit ourselves to api related only

BanulaKumarage: > @BanulaKumarage Yeah hosting swagger specs directly where the site is hosted is fine. I remember seeing this when we did the last implementation of drf-yasg, we saw some issues with the generation of documentation for fields in the api. Can you check if every api has defined doc for every field? The actual documentation is currently hosted here - https://github.com/spdx/spdx-online-tools/wiki/REST-API-Fields-Request-and-Response - but is hard to find and also hard to track and update. Integrating directly into the website as a separate page looks perfect.

Hi @rtgdk I made this as an initial step I will check every field. And update you regarding this

BanulaKumarage: > @BanulaKumarage Also, looks like some extra apis are generated as well. We should limit ourselves to api related only

@rtgdk  is it the auth apis? which are the extra ones? There should be a way to avoid those. I will try to make it clean.

rtgdk: @BanulaKumarage Yes, auth apis and `api2` seems weird. Can you confirm if these are even needed for swagger specs?

BanulaKumarage: @rtgdk parameters are available in api2 but not in API endpoints. I have tried multiple libraries but I got the same output. I am trying to figure out a way of resolving this.

bact:
Close PRs #323 (drf-yasg for OpenAPI 2 schema + Swagger/Redoc UI) and #451 (Swagger) as they are superseded by #644 (drf-spectacular for OpenAPI 3 schema + Swagger/Redoc UI).

Feel free to open new PR to amend/extend the API doc generation.

Issue #398, opened 2022-10-15
License submission form-XML editor: exception v license


jlovejoy:
XML creator automatically uses "license" even if it's an exception. Not sure there's a way to change this as we'd probably need a way to designate it as an exception in the submission form?

goneall:
@jlovejoy Agree it would require some changes to the UI.

jlovejoy:
Some more notes on this:
- add tick box as to license or exception in submission form
- this then dictates the appropriate tag when creating the XML and the PR (to put the XML in the exceptions folder)

Also note: currently, if you change License --> exception in that tag, and then "beautify" it strips out the other attributions for the license/exception tag.

HirumalPriyashan:
Hi @jlovejoy,

> Also note: currently, if you change License --> exception in that tag, and then "beautify" it strips out the other attributions for the license/exception tag.

Can you kindly explain what is meant by this line?

jlovejoy:
it's hard to explain in text, if you test the XML editor via one of the submissions it will make more sense. Basically, if you try to change the license tag to exception. then use the "beautify" function in the xml editor, it "corrects" the tag back to <license>

HirumalPriyashan:
I see, thanks for the clarification.
I noticed `beautify` strips out the content of the exception tag in that case.

jlovejoy:
@goneall @rtgdk - looks like there are two PRs open against this issue - should we close one?

jlovejoy:
@BassCoder2808 - could you comment here (so I can assign you to this issue) :)

BassCoder2808:
Sure @jlovejoy

jlovejoy:
ah, the related PRs to this issue have not been merged, so won't see fix yet!

bact:
Is this one already fixed by https://github.com/spdx/spdx-online-tools/pull/496 ?

goneall:
Looks like it - I'll go ahead and close this issue.

PR #458, opened 2023-04-01
Fixing the issue of striping attributes when beautify

BanulaKumarage:
- As mentioned in #398 attributes of the tag were missed when changing the tag from license to exception and beautify in the XML editor. 
- It was fixed by updating the ATTRS_SEQ list in `formatlXml.py`

### DIFF
--- a/src/app/formatxml.py
+++ b/src/app/formatxml.py
@@ -62,6 +62,11 @@ ATTRS_SEQ = {
             'name',
             'listVersionAdded',
         ],
+        'exception': [
+            'licenseId',
+            'name',
+            'listVersionAdded',
+        ],
         'alt': [
             'name',
             'match',
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,58 @@
-This project tracks patch provenance and licensing using the
-[Developer Certificate of Origin][DCO] and Signed-off-by tags
-initially developed by the Linux kernel project.  Because the
-documentation for this procedure is licensed under the GPLv2, we have
-chosen not to include it in our project directly.  Instead, please see
-[SubmittingPatches][], which is stored in an external repository.
+# Contributing to SPDX Online Tools
 
-[DCO]: developer-certificate-of-origin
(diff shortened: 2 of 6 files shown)


BanulaKumarage:
@jlovejoy Can you review this?

jlovejoy:
@rtgdk - I'm not sure how the functionality for the "beautify" option works or if this is all that is needed to fix this issue. Can you please review??

src/app/formatxml.py
```diff
@@ -13,6 +13,7 @@
 
 INDENT = 2
 LINE_LENGTH = 80
+IS_EXCEPTION = False
```
rtgdk, line 16: Having a global variable can affect other users as well since the global value is changed. The application won't have separate instances of `app` for every user.

src/app/formatxml.py
```diff
@@ -143,7 +152,9 @@ def pretty(node, level):
         text = text.replace('&', '&amp;').replace('>', '&gt;').replace('<', '&lt;')
         ser += text
     for child in node:
-        ser += pretty(child, child_level)
+        res = pretty(child, child_level)
+        if not(IS_EXCEPTION and 'standardLicenseHeader' in res):
```
rtgdk, line 156: This needs to checked in the context of `pretty` func not at a global level I think.

rtgdk (commented):
Left some comments

src/app/formatxml.py
BanulaKumarage: Done

src/app/formatxml.py
```diff
@@ -143,7 +148,9 @@ def pretty(node, level):
         text = text.replace('&', '&amp;').replace('>', '&gt;').replace('<', '&lt;')
         ser += text
     for child in node:
-        ser += pretty(child, child_level)
+        res = pretty(child, child_level)
+        if not('exception' in ser and 'standardLicenseHeader' in res):
```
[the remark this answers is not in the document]
BanulaKumarage: @sonatype-lift ignore

goneall:
@BanulaKumarage @rtgdk - Should this be merged?

Issue #455, opened 2023-03-30
Design Issue on Submit License Page


BassCoder2808:
On the page [Submit New License](https://tools.spdx.org/app/submit_new_license/), there is a patch of white space between the above toolbar and the start of the textbox as we can see in the image. This should not be there as it is a wastage of space as well as not solving anything
![image](https://user-images.githubusercontent.com/65075935/228808688-86bcbacd-3cee-46a9-a67e-bb643a805425.png)

BassCoder2808:
Upon further inspection, the reason why this is occurring is that we have intentionally left space for the modal messages above, instead what we can do is that we will need to change the way we handle the modal messages in the `base.html` which will automatically solve this issue
![image](https://user-images.githubusercontent.com/65075935/229279045-d91484f4-599a-49cc-93a8-b5a472c808ad.png)

BassCoder2808:
I have modified this and the new page would look something like this
![image](https://user-images.githubusercontent.com/65075935/229334375-caea4766-aeaf-48c3-81d0-3d5baa7e39d6.png)

goneall:
Resolved with PR #459

PR #459, opened 2023-04-02
Intended to Solve the design issue on the Submit License Page

BassCoder2808:
- Intended to solve #455 
- Created a new base file for the submit license page so that the structure of other files is not disturbed
- Removed `{mediaLink}` and `{error}` as they are not used in the templates and is part of the `API`
- Changed the location of the `messages` div


BassCoder2808:
@jlovejoy and @rtgdk can you please review this if something else needs to be done in this

src/app/templates/app/submit_new_license_base.html
```diff
@@ -0,0 +1,196 @@
+<!DOCTYPE html>
```
rtgdk, line 1: @BassCoder2808 
Having a new base file for submit license is not needed, this should be fixed in the base.html itself. 

Reasons:-
1. The whole website should have only one base file.
2. Maintenance is difficult if we have the same content in multiple files.
3. Website would be loading a new html page specifically for submit license page - which is not required.


Let's try to figure out how to fix this globally in base.html

BassCoder2808: Sure @rtgdk , I will try to figure this out and will let you know once thats done

src/app/templates/app/submit_new_license.html
```diff
@@ -4,12 +4,15 @@
...
+    <style>
+      .body1 {
+        display: none;
+      }
+    </style>
 {% endblock %}
 
-{% block body1 %}
-    {% include "app/modal.html" %}
-{% endblock %}
```
rtgdk, line 11: Can't remove modal.html. This modal box is being used in this html page for github login

rtgdk, line 11: Can't remove modal.html. This modal box is being used in this html page for github login

src/app/templates/app/base.html
```diff
@@ -181,7 +181,7 @@
     </div><!--/.nav-collapse -->
   </nav>
 
-    <div class="container">
+    <div class="container body1">
```
rtgdk, line 184: I see this in base.html
```
<div class="container">

      <div class="starter-template">
        {% block body1 %}{% endblock %}
      </div>

    </div><!-- /.container -->
		{% block body2 %}{% endblock %}
```

submit_new_license is made with body2. Can you try moving it to body1 coz body1 will always be present and take up some space. For me deleting the whole container element on this page gave me least space. If we use `body1` block itself for this page then it would be similar to other pages.

rtgdk:
@BassCoder2808 Any update on this?

src/app/templates/app/submit_new_license.html
```diff
@@ -54,13 +57,11 @@
 {% else %}
 <p id="githubLogin" hidden>False</p>
 {% endif %}
-<div id="messages" class="messages">
 </div>
-<p class ="lead"> {{ medialink }}</p>
-<p class ="lead"> {{ error }}</p>
```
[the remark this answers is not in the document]
BassCoder2808: Yes the error is directly send to the `modal` and this is not required

src/app/templates/app/base.html
BassCoder2808: Hi @rtgdk , I tried putting the page itself in `block1` but the problem is that the block one is specifically only being used for the modal, because of which it doesn't take up the full space and instead the page looks like
image
So in the current code I have made use of `block2` let me know if anything in that needs to be changed

Issue #376, opened 2022-06-09
Validating an SPDX file with colon in name fails


vargenau:
In the online validator, I try to validate a file called `mariadb:latest.spdx`

I get a banner with the following error message: 

```
File /spdx/src/app/media/AnonymousUser/1654771839/mariadb%3Alatest.spdx not found.
```

rtgdk:
@vargenau Which OS are you on when you upload this file? colon are not prohibited in windows world. Can you try without colon filename?

vargenau:
It was a Linux machine, Ubuntu.

The bug is that the colon ":" has been encoded as "%3A". It should not have been.

BanulaKumarage:
Hi, @rtgdk I am working on a windows environment. As you said, colons are prohibited in windows. Does this error occur in other characters except for the colon? If so I can try with that to reproduce the error. Otherwise is there any other way to reproduce this error? Please assist me on this.

rtgdk:
@BanulaKumarage You can try to download a file which has colon in it's name and try opening it in Windows or through the onlint tool. You should be able to reproduce it.

There should a way to escape colon character or rename it before opening in windows.

BanulaKumarage:
Thanks @rtgdk I will try that

BanulaKumarage:
> @BanulaKumarage You can try to download a file which has colon in it's name and try opening it in Windows or through the onlint tool. You should be able to reproduce it.
> 
> There should a way to escape colon character or rename it before opening in windows.

when I tried to download such a file, it automatically renamed the file as follows.

![image](https://user-images.githubusercontent.com/63106638/229342293-2e0ca7f6-6b28-42df-84af-2fb877ca52e1.png)

BanulaKumarage:
![image](https://user-images.githubusercontent.com/63106638/229349927-4872f923-e1b8-4edb-9218-f90fd907f117.png)

Recreated the issue. Working on it

BanulaKumarage:
@rtgdk @vargenau Made PR #460 to fix this. Please review.

vargenau:
Hi @BanulaKumarage 
I cannot test if the fix is not deployed.

vargenau:
Has the fix been deployed on https://tools.spdx.org/app/validate/ ?

I still have the error.

BanulaKumarage:
> Has the fix been deployed on https://tools.spdx.org/app/validate/ ?
> 
> I still have the error.
> 

Its not merged it yet.

[...]

PR #460, opened 2023-04-02
Fixing #376

BanulaKumarage:
- Earlier file names with non-ASCII characters were converted to ASCII characters in `fs.url(filename)` line 321 in `app/core.py` But the file is saved without converting in line 320. Therefore, the file is unable to find.
- As a fix, the file is saved in `app/media` by converting its non-ASCII characters to ASCII characters. Then issue #376 is fixed.

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -6,7 +6,6 @@ from django.http.response import JsonResponse
 import jpype
 import os
 import sys
-import unicodedata
 
 from django.utils.datastructures import MultiValueDictKeyError
 from django.core.files.storage import FileSystemStorage
@@ -317,7 +316,7 @@ def license_validate_helper(request):
             fs = FileSystemStorage(location=settings.MEDIA_ROOT +"/"+ folder,
                 base_url=urljoin(settings.MEDIA_URL, folder+'/')
                 )
-            filename = fs.save(unicodedata.normalize('NFKD', myfile.name).encode('ASCII', 'ignore').decode('ASCII'), myfile)
+            filename = fs.save(utils.removeSpecialCharacters(myfile.name), myfile)
             uploaded_file_url = fs.url(filename).replace("%20", " ")
             formatstr = request.POST["format"]
             serFileTypeEnum = jpype.JClass("org.spdx.tools.SpdxToolsHelper$SerFileType")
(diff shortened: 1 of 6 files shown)


src/app/core.py
```diff
@@ -316,7 +317,7 @@ def license_validate_helper(request):
             fs = FileSystemStorage(location=settings.MEDIA_ROOT +"/"+ folder,
                 base_url=urljoin(settings.MEDIA_URL, folder+'/')
                 )
-            filename = fs.save(myfile.name, myfile)
+            filename = fs.save(unicodedata.normalize('NFKD', myfile.name).encode('ASCII', 'ignore').decode('ASCII'), myfile)
```
rtgdk, line 320: @BanulaKumarage Can you please explain the changes. There seems to be a lot of encode/decode happening. Since this is the main flow, don't want to merge until I understand what exactly it does.

Also, can you create a util to do this since this would be required at other places as well.

Also, having some UT for this would be good.

rtgdk:
Update PR heading as well.

src/app/core.py
BanulaKumarage: `unicodedata.normalize` method, converts any special character into their equivalent base characters or combinations of base characters. (ex: converting _é_ to _e_ ) Then the text is converted to ASCII characters. But if there is any character cannot convert it is replaced with a empty string( done by `ignore`) 
Finally, the resulting ASCII-encoded string is decoded back into a regular Python string, which can then be used as the filename. ( converting `b'mariadblatest.spdx'` to `mariadblatest.spdx` )

src/app/utils.py
```diff
@@ -367,6 +368,10 @@ def postToGithub(message, encodedContent, filename):
     return r.status_code, r.json()
 
 
+def removeSpecialCharacters(filename):
+    return unicodedata.normalize('NFKD', filename).encode('ASCII', 'ignore').decode('ASCII')
```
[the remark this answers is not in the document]
BanulaKumarage: @sonatype-lift ignoreall

rtgdk:
This doesn't seem to work for me on Mac - Upload
Screenshot 2023-06-29 at 5 53 14 PM

Online tools also failing with colon in filename with these changes
Screenshot 2023-06-29 at 6 33 57 PM

I think it's better to substitute all illegal filename characters instead, will push the changes.

PR #471, opened 2023-05-30
Differentiate Exceptions from License Submission

BassCoder2808:
- Intended to solve #398 
- Added the new fields in the `models.py`
- Changed the `formatXML.py` to solve the issue 
- Added a new field in `form.py` to accommodate the `isException` field


jlovejoy:
@BassCoder2808 -  a couple comments / suggestions:
- the `isOsiApproved` tag is not needed/relevant for exceptions, neither is `standardLicenseHeader` so it would be good if the generated XML could omit those, if possible
- in the submission form - will the submitter just see "Is Exception" and a text box to fill in? I think we could have that be a yes/no option?
- we probably ought to also update the explanatory text to include "license or exception" on most of the fields. happy to help with that part

src/app/generateXml.py
```diff
@@ -140,7 +140,10 @@ def generateLicenseXml(licenseOsi, licenseIdentifier, licenseName, listVersionAd
         licenseOsi = "true"
     else:
         licenseOsi = "false"
-    license = ET.SubElement(root, "license", isOsiApproved=licenseOsi, licenseId=licenseIdentifier, listVersionAdded=listVersionAdded, name=licenseName)
+    if isException:
+        license = ET.SubElement(root, "exception", isOsiApproved=licenseOsi, licenseId=licenseIdentifier, listVersionAdded=listVersionAdded, name=licenseName)
```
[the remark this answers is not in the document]
BassCoder2808: @sonatype-lift ignoreall

BassCoder2808:
Hi @jlovejoy for your comments, yes I will remove the `isOsiApproved` tag for the exception part and for the `standardLicenseHeader` it is automatically added when a user puts that so that shouldn't be a problem for the `exception` XML generation.

> * we probably ought to also update the explanatory text to include "license or exception" on most of the fields. happy to help with that part
This would be really helpful if you could help me with this @jlovejoy, till that time I will finish up with adding the tooltip then we can change the text inside the tooltip

src/app/utils.py
```diff
@@ -171,26 +171,54 @@ def makePullRequest(username, token, branchName, updateUpstream, fileName, commi
...
+        file_url = "{0}/contents/src/{1}".format(TYPE_TO_URL_NAMESPACE[NORMAL] if is_ns else TYPE_TO_URL_LICENSE[NORMAL], fileName)      
     response = requests.get(file_url, headers=headers)
+    text_file_url = "{0}/contents/test/simpleTestForGenerator/{1}".format(TYPE_TO_URL_NAMESPACE[NORMAL] if is_ns else TYPE_TO_URL_LICENSE[NORMAL], textFileName)
+    text_response = requests.get(text_file_url, headers=headers)
     if response.status_code == 200:
         """ Creating Commit by updating the file """
         data = json.loads(response.text)
         file_sha = data["sha"]
         body["sha"] = file_sha
+    if text_response.status_code == 200:
```
rtgdk, line 214: Some changes looks similar to https://github.com/spdx/spdx-online-tools/pull/470/ . Please look into working with multiple branches on git. Create PRs with the correct upstream branch (branch before this one)

src/app/migrations/0002_auto_20230529_1433.py
```diff
@@ -0,0 +1,33 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='licensenamespace',
+            name='isException',
+            field=models.BooleanField(default=True),
+        ),
+        migrations.AddField(
+            model_name='licensenamespace',
+            name='text',
+            field=models.TextField(default=''),
```
rtgdk, line 21: You will need to commit only changed migrations if this is fixed in the other PR

rtgdk (commented):
Left some comments

BassCoder2808:
Hi @rtgdk , I have created a new pull request #492 because lately there has been some issues with the migration files, this was also experienced in #470.
I have made all the changes which you had suggest in #492, let me know if anything else needs to be done in that

src/app/utils.py
BassCoder2808: I have made a new PR #492 for this, like further moving forward I have to redo #470 because of the migrations issues which it has having, so I have incorporated all the changes which you had suggested in #470 in #492, so now there is an extra utils function as well. Let me know if anything else needs to be done in that

Issue #494, opened 2023-07-19
HUGE warning for deprecated licences


vargenau:
The following SPDX file has:
one line with `PackageLicenseDeclared: LGPL-2.1`
two lines with `PackageLicenseDeclared: LGPL-2.1+`

[airflow-tern2.12.1.spdx.txt](https://github.com/spdx/spdx-online-tools/files/12093762/airflow-tern2.12.1.spdx.txt)

So we expect warnings about deprecated licenses.

The issue is that the warning message is HUGE and mentions relationships.

Extract:
```
[...]

PR #495, opened 2023-07-26
Intended to solve #494

BassCoder2808:
- Made the changes suggested by @vargenau and @goneall which will help better format the error messages and warnings we receive in the online tools.
-  Solves #494

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -222,7 +222,7 @@ def ntia_check_helper(request):
                 if (request.is_ajax()):
                     ajaxdict["type"] = "warning"
                     warnings = str(retval)
-                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />')
+                    ajaxdict["data"] = "The following warning(s) were raised:<br />\n" + warnings.replace('\n', '<br />\n')
                     response = dumps(ajaxdict)
                     result['response'] = response
                     result['status'] = 400
@@ -329,7 +329,7 @@ def license_validate_helper(request):
                 if (request.is_ajax()):
                     ajaxdict["type"] = "warning"
                     warnings = str(retval)
-                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />')
+                    ajaxdict["data"] = "The following warning(s) were raised:<br />\n" + warnings.replace('\n', '<br />\n')
                     response = dumps(ajaxdict)
                     result['response'] = response
                     result['status'] = 400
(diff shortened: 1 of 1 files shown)


vargenau:
Hi @BassCoder2808,
Did you test your changes?
In the third case, I suppose you should have:
```
warnings = str(warnings)
```

BassCoder2808:
Hi @vargenau yes I just saw that the 3rd case had the `warnings` string, now I have made that change

vargenau:
> Hi @vargenau yes I just saw that the 3rd case had the `warnings` string, now I have made that change

Thank you @BassCoder2808

src/app/core.py
```diff
@@ -221,7 +221,8 @@ def ntia_check_helper(request):
                 """ If any warnings are returned """
                 if (request.is_ajax()):
                     ajaxdict["type"] = "warning"
-                    ajaxdict["data"] = "The following warning(s) were raised:\n" + str(retval)
+                    warnings = str(retval)
+                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />')
```
goneall, line 225: Just a nit - it would be nice to keep a linefeed in case anyone looks at the raw HTML.

goneall, line 225:
```suggestion
                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />\n')
```

src/app/core.py
```diff
@@ -327,7 +328,8 @@ def license_validate_helper(request):
                 """ If any warnings are returned """
                 if (request.is_ajax()):
                     ajaxdict["type"] = "warning"
-                    ajaxdict["data"] = "The following warning(s) were raised: " + str(retval)
+                    warnings = str(retval)
+                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />')
```
goneall, line 332:
```suggestion
                    ajaxdict["data"] = "The following warning(s) were raised:\n" + warnings.replace('\n', '<br />\n')
```

src/app/core.py
```diff
@@ -519,7 +521,8 @@ def license_convert_helper(request):
             else :
                 if (request.is_ajax()):
                     ajaxdict["type"] = "warning"
-                    ajaxdict["data"] = "The following warning(s) were raised by "+ myfile.name + ": " + str(warnings)
+                    warnings = str(warnings)
+                    ajaxdict["data"] = "The following warning(s) were raised by "+ myfile.name + ": " + warnings.replace('\n', '<br />')
```
goneall, line 525:
```suggestion
                    ajaxdict["data"] = "The following warning(s) were raised by "+ myfile.name + ": " + warnings.replace('\n', '<br />\n')
```

src/app/core.py
vargenau: We could also have
```
"The following warning(s) were raised:<br />\n"
```
instead of:
```
"The following warning(s) were raised:\n"
```
so that the first warning is not on the same line.

BassCoder2808:
> Thanks @BassCoder2808 - just a minor suggested change to format the raw HTML a bit better

Hi @goneall I have made the changes, let me know if anything else needs to be changed

vargenau (approved):

goneall (approved):
Thanks @BassCoder2808  and @vargenau

PR #497, opened 2023-08-10
Created Folder structure for tests for app

BassCoder2808:
- Created a folder `tests`
- Segregates all the class tests into their respective files for easier test handling


BassCoder2808:
Hi @rtgdk can you please check this once and let me know if I am doing it correctly, like we need to do something similar for the `api` folder so, if this one looks good, I will follow the similar structure for it as well

src/app/tests/archieve_license_namespace_views_test.py
```diff
@@ -0,0 +1,98 @@
...
+from webdriver_manager.firefox import GeckoDriverManager
+
+from app.models import LicenseNamespace
+from app.generateXml import generateLicenseXml
+
+
+
+class ArchiveLicenseNamespaceViewsTestCase(StaticLiveServerTestCase):
+
+    def setUp(self):
```
rtgdk, line 18: Let's have a common setup and teardown function in `test_utils.py` and use it wherever we can.

rtgdk (commented):
Please provide the screenshot and the command used to run all test cases. Show the output as successfull

BassCoder2808:
> Please provide the screenshot and the command used to run all test cases. Show the output as successfull

image
Hi @rtgdk I think all the tests are working like I tried this on both the main and `Test-Files` branch and the tests which are failing are common like they are the firefox ones I think it cannot find the relevant webdriver on my local wsl setup
So once I like do the same for the other modules can you please run my branch locally before merging it

rtgdk:
Overall changes LGTM. I hope these is no code leakage in this - very difficult to review for that. I would have loved to have a common setup/teardown for all tests and having all tests run without failure/skipping but I guess these will need some rethinking on how to run test cases run - currently a lot of duplicate code exists .

@BassCoder2808 Can you please list down the test cases not working on your machine?  This will make the dataset for tests affected by selenium/firefox usecases and provide us a start to move these to unit based framework.

There are already some issues opened for fixing these testcases. Let's try to fix some of them.

Holding to merge this diff for sometime - until I run tests on this branch locally.

bact:
If we can have this separated, it will be possible that during Docker image build we can exclude all these tests for the production image - to reduce image size.

bact:
Since the tests here are almost 3 years. Few of them has changes, so it needs more review if we like to merge this.

bact:
I will rework on this, based on @BassCoder2808 work and @rtgdk feedback.

Plan to do the app tests first, as a group of smaller PRs to make the review easier.
Once finished the app tests, will do the same for API tests.

bact:
I have started the work. The 1st PR of the series is at https://github.com/spdx/spdx-online-tools/pull/684

bact:
> If we can have this separated, it will be possible that during Docker image build we can exclude all these tests for the production image - to reduce image size.

PR #693 will move all API and App tests + fixtures to be under one tests/ root. 

It also added that tests/ root to .dockerignore.

bact:
With https://github.com/spdx/spdx-online-tools/issues/697 closed as completed, we can close this issue now. Thank you @BassCoder2808 again who initiated this work.

PR #500, opened 2023-08-31
Added the GUI changes

BassCoder2808:
- Added the GUI changes for the license and exceptions as discussed in the doc shared by @jlovejoy 
@rtgdk can you review this once and let me know if anything else needs to be changed in this

### DIFF
--- a/src/app/templates/app/submit_new_license.html
+++ b/src/app/templates/app/submit_new_license.html
@@ -70,7 +70,7 @@ Once you have submitted a license, please follow the <a href="https://github.com
 
     <div class = "form-group">
       <div class="col-sm-12">
-      <label class="control-label col-sm-3" >Briefly explain how this license <br />meets the
+      <label class="control-label col-sm-3" >Briefly explain how this license or exception<br />meets the
         <a href="https://github.com/spdx/license-list-XML/blob/main/DOCS/license-inclusion-principles.md">
           license inclusion principles
         </a>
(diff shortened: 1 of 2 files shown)


rtgdk:
@BassCoder2808 Please add the screenshots for this change.
@jlovejoy Can you please review the screenshot and confirm the changes you wanted?

src/app/templates/app/submit_new_license.html
```diff
@@ -70,7 +70,7 @@
 
     <div class = "form-group">
       <div class="col-sm-12">
-      <label class="control-label col-sm-3" >Briefly explain how this license <br />meets the
+      <label class="control-label col-sm-3" >Briefly explain how this license or exception<br />meets the
```
rtgdk, line 73: Isn't it better to have either shown license or exception?
Like if the dropdown selected is license then we show only `Briefly explain how this license` and when selected exception - `Briefly explain how this exception`. 
@jlovejoy Let me know your thoughts. I think since the dropdown option comes later in the form, it might get confusing.

@BassCoder2808 Please do the needful if needed. Overall LGTM. Will merge after @jlovejoy approvals.

rtgdk:
Posting the screenshot from my local testing. 

Screenshot 2023-10-03 at 6 19 44 AM
@BassCoder2808 I think you covered all the points mentioned by @jlovejoy on the doc from the UI text change perspective. Merging this now as it LGTM.

jlovejoy:
not sure what the above comment was referencing, but looking at the screenshot, I'm assuming the "is this a license exception" defaults to no, and the drop-down allows one to choose yes - if so, that seems good!

rtgdk:
@jlovejoy Yes that is true.

PR #507, opened 2023-10-03
Small code maintenance fixes

rtgdk:
- Small fixes for code maintenance 
- Renamed some variable for better readability
- Fixed matching with true value


goneall (approved):
LGTM

Issue #512, opened 2023-10-19
"Submit license" form apparently only allows 25 characters for suggested short identifier


richardfontana:
While one might reasonably think that, in coming up with a short identifier for a license, "25 characters ought to be enough for anybody", the growing complexity of the SPDX license list (including relationships between identifiers) probably makes this limitation inadequate. I am trying to submit a license that is a variant of `FSFAP` but without the final sentence containing a simple warranty disclaimer. So a natural choice of short identifier is `FSFAP-no-warranty-disclaimer` but this exceeds the apparent 25 character limit.

goneall:
@richardfontana - should we have a limit on the length of the ID's?  If so, what would be reasonable?

richardfontana:
@goneall my thought was perhaps double the existing limit?

richardfontana:
I was again hit by this limitation when attempting to submit a new license. I wanted to suggest the identifier `BSD-3-Clause-Intel-variant` but the form would only give space for `BSD-3-Clause-Intel-varian`.

goneall:
Due to the demands from the SPDX 3.0 release, I probably won't get to a fix in the near term.  Hoping to find a volunteer in the community to provide a pull request.

Ash-KODES:
Can I work on this ? @goneall

goneall:
@Ash-KODES - that would be great - thanks!

goneall:
@Ash-KODES It would be great to get in a pull request for this soon - I'm planning on doing a release this week.

Let me know if you're able to take this on - otherwise I'll take a stab at it.

cc: @rtgdk

Ash-KODES:
I will make a PR for this  today

goneall:
@Ash-KODES - Just checking in on the PR - let me know if you think you'll be able to get the PR done in the next day or so

BassCoder2808:
Hi @goneall, If the issue is still open I would like to contribute to it, I will create the PR for the same and you can review it and we can move forward with it

BassCoder2808:
We can close this issue now @goneall

PR #515, opened 2024-01-02
Changed the length of the shortIdentifier field

BassCoder2808:
- Intended to solve #512 
- Changed the length of the field from **25** to **50**

@goneall please let me know if anything else needs to be done in this


goneall (approved):
LGTM

Thanks @BassCoder2808

Issue #511, opened 2023-10-11
Check license no longer works


goneall:
After the upgrade, when submitting a new license - if you click on check license, nothing happens.

The browser console log has the following error:

```
(index):896 Uncaught ReferenceError: findLicenseMatch is not defined
    at HTMLButtonElement.<anonymous> ((index):896:5)
```

goneall:
@BassCoder2808 @rtgdk - could you take a look at this

BassCoder2808:
Hi @goneall, I checked this and have attached the SS below, it is working on my local machine, does the issue still persist?
because the  

> findLicenseMatch
function is present in the utils.js which is loaded on line 177 in submit_new_license.html

![image](https://github.com/spdx/spdx-online-tools/assets/65075935/4fe7f25f-bbc7-4e26-a73e-14de6d20ca45)

goneall:
@BassCoder2808 I just checked the deployed system at https://tools.spdx.org/app/submit_new_license/ and the error is still there.

It looks like the function is defined in the file `static/js/util.js` but that file is not present on the deployed system - see below.

Is it possible there is some deployment issue where that file isn't being deployed?

![image](https://github.com/spdx/spdx-online-tools/assets/378172/e1aa94d8-41cb-411d-93c6-0d0c4f1e92fa)

BassCoder2808:
I think the problem is that we are not running the collect static command and the STATIC_ROOT URL is not defined

> Static File Configuration: In your Django settings, ensure that the STATIC_ROOT and STATIC_URL are correctly configured for production. STATIC_ROOT should point to where your static files are collected for deployment.

> Collect Static Files: Run python manage.py collectstatic to gather all static files from your apps into one location. Ensure this command is run before deploying your Django app.

> Web Server Configuration: Check if your web server (like Nginx or Apache) is correctly configured to serve static files. It should be configured to serve files from the STATIC_ROOT directory.

> Permissions: Verify that the user running the web server process has permissions to access the static files in the deployed directory.

> Debug Mode: If Django is in debug mode (DEBUG = True), static files might be served by Django itself, which could differ from how they are served in production. Ensure DEBUG is set to False in the production environment.

> Cache Busting: Check if there's any caching mechanism interfering with the visibility of the updated static files. Clearing the cache or using cache-busting techniques might resolve the issue.

> Deployment Scripts: Review your deployment scripts to ensure they are correctly copying static files to the appropriate location in your production environment.

[...]

PR #516, opened 2024-01-04
Collect Static

BassCoder2808:
- Intended to solve #511 
- Made changes to `settings.py` file to add `STATIC_ROOT`
- Ran `python manage.py collecstatic`


goneall:
Thanks @BassCoder2808 for pulling this together.

It looks like the script added a lot of files that don't currently have an issue.

I'm not very familiar with Django to know if this is expected.

@rtgdk - Could you take a look and review? You know Django much better than I do.

BassCoder2808:
Hi @goneall, was going through the code and a better strategy which I can currently think of is to run the `collectstatic` command in our dockerfile, this will reduce the files modified as well and only 2 files will get changed, will create a new PR for the same

PR #518, opened 2024-01-05
Added STATIC_ROOT and changed docker.dev

BassCoder2808:
- Intended to solve #511 

@goneall have changed `dockerfile.dev` and added the `collectstaic` command there. This should work and all the static files should get collected. Can you try that once and do let me know if anything else needs to be done in this.

### DIFF
--- a/dockerfile.dev
+++ b/dockerfile.dev
@@ -29,5 +29,6 @@ EXPOSE 8000
 
 # Run migrations and start the development server on port 8000
 CMD python src/manage.py migrate && \ 
+    python src/manage.py collectstatic && \ 
     python src/populate.py && \
     python src/manage.py runserver 0.0.0.0:8000
--- a/src/src/settings.py
+++ b/src/src/settings.py
@@ -19,7 +19,7 @@ APP_DIR = os.path.join(BASE_DIR,'app')
 API_DIR = os.path.join(BASE_DIR,'api')
 TEMPLATE_DIR = os.path.join(APP_DIR, 'templates')
 STATIC_PATH = os.path.join(APP_DIR,'static')
-STATIC_ROOT = os.path.join(APP_DIR,'static')
+STATIC_ROOT = STATIC_PATH
 EXAMPLES_DIR = os.path.join(BASE_DIR,'examples')
 
 LICENSE_PROD_REPO_NAME = "license-list-XML"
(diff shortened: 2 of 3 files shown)


dockerfile.dev
```diff
@@ -29,5 +29,6 @@ EXPOSE 8000
 
 # Run migrations and start the development server on port 8000
 CMD python src/manage.py migrate && \ 
+    python src/manage.py collectstatic && \
```
goneall, line 32: Please make the same change to the dockerfile.prod

goneall (changes requested):
Thanks @BassCoder2808 - I like this approach.  Just one change requested.

dockerfile.dev
BassCoder2808: @goneall was going through the `dockerfile.prod` but couldn't find how we are starting the django server.
wanted to ask like do we manually run the commands while deploying or how is it?

src/src/settings.py
```diff
@@ -19,6 +19,7 @@
 API_DIR = os.path.join(BASE_DIR,'api')
 TEMPLATE_DIR = os.path.join(APP_DIR, 'templates')
 STATIC_PATH = os.path.join(APP_DIR,'static')
+STATIC_ROOT = os.path.join(APP_DIR,'static')
```
rtgdk, line 22: Let's make it 
```
STATIC_ROOT = STATIC_PATH
```

rtgdk (changes requested):

dockerfile.dev
goneall: It is run under [supervisor](http://supervisord.org/).  The last line in the docker file starts it up.

[supervisor_api.conf](https://github.com/spdx/spdx-online-tools/blob/main/supervisor_api.conf) is the config file.

Supervisor starts up [gunicorn](https://gunicorn.org/) to run a [wsgi](https://wsgi.readthedocs.io/en/latest/what.html) script which run Django via wsgi.

Here's some background on how that is setup: https://docs.djangoproject.com/en/5.0/howto/deployment/wsgi/

BassCoder2808: Thanks @goneall for this, I will make the appropriate changes now

rtgdk (approved):

goneall (approved):
Updates look good - Thanks @BassCoder2808

PR #523, opened 2024-01-19
Add --noinput to static file generation

goneall:
Resolves an issue that if there are existing files, there is a prompt to overwrite which causes the docker build to fail


BassCoder2808:
LGTM @goneall

PR #524, opened 2024-01-19
Add static generated files

goneall:
I'm not sure if these files should be added to the repo.

If not, we should modify the `.gitignore` file so they will be ignored.

These were generated when testing out the latest build using `dockerfile.dev`.

@BassCoder2808 @rtgdk - thoughts?


BassCoder2808:
Hi @goneall, I think yes we should add these files as they are static content, but it can be skipped as well and in `README.md` we can add the collect static command if somebody wants to run or test it

goneall:
Thanks @BassCoder2808 - I'll go ahead and merge this in

Issue #530, opened 2024-04-07
Fix all api and app Unit Tests to run successfully on all environments


rtgdk:
Idea is to make these tests run successfully and smoothly on any platform.
They should be easily runnable via PyCharm and in docker container as well.
Github workflow should also run these in consistent manner

bact:
With https://github.com/spdx/spdx-online-tools/pull/628 merged, [.github/workflows/test.yml](https://github.com/spdx/spdx-online-tools/blob/main/.github/workflows/test.yml) is now triggered and run for every PR and push. See for example,

- https://github.com/spdx/spdx-online-tools/actions/runs/24972444940
- https://github.com/spdx/spdx-online-tools/actions/runs/24966282372

I think we can close this issue as fixed.

bact:
Since the tests (`python src/manage.py test api app`) are run smoothly in the past two months. Both locally (macOS) and on GitHub CI (Ubuntu - most recent run few minutes ago: https://github.com/spdx/spdx-online-tools/actions/runs/28022193954/job/82940994677?pr=679). I will close this issue.

Please reopen if any test failed on any platform.

PR #531, opened 2024-04-07
Fix test_checklicense_api 

rtgdk:
Fixing #530 

Had some free time, trying to make all tests run successfully - both locally and via PyCharm.
PyCharm debugger is very helpful to fix these issues faster.
Will be fixing more tests in app/tests.py as well.

- Fixed test_checklicense_api where response structure changed
- Fixed core.py with proper exception handling

Now only 2 tests in api are failing - both with rdf file 
Screenshot 2024-04-07 at 2 56 53 PM

@goneall Has there been any change with verifying RDF files. Both the remaining tests are failing as an exception is raised by `verifyclass.verify` JVM call to verify RDF files - `'org.spdx.library.model.SpdxIdInUseException: Can not create Apache-2.0.  It is already in use with type License which is incompatible with type ListedLicense')`

### DIFF
--- a/src/api/tests.py
+++ b/src/api/tests.py
@@ -56,7 +56,7 @@ class ValidateFileUploadTests(APITestCase):
         u.is_staff = True
         u.save()
         self.tv_file = open(getExamplePath("SPDXTagExample-v2.0.spdx"))
-        self.rdf_file = open(getExamplePath("SPDXRdfExample-v2.0.rdf"))
+        self.rdf_file = open(getExamplePath("SPDXRdfExample-v2.2.spdx.rdf"))
         self.invalid_tv_file = open(getExamplePath("SPDXTagExample-v2.0_invalid.spdx"))
         self.invalid_rdf_file = open(getExamplePath("SPDXRdfExample-v2.0_invalid.rdf"))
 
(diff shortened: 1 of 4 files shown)


rtgdk:
@goneall Is there a better RDF file we can use for tests? The current one - https://github.com/spdx/spdx-online-tools/blob/main/src/examples/SPDXRdfExample-v2.0.rdf - raise issues like `'org.spdx.library.model.SpdxIdInUseException: Can not create Apache-2.0.  It is already in use with type License which is incompatible with type ListedLicense')`. Tried with  this 2.1 version as well but same issue - https://github.com/spdx/tools/blob/master/Examples/SPDXRdfExample-v2.1.rdf

goneall:
> @goneall Is there a better RDF file we can use for tests? The current one - https://github.com/spdx/spdx-online-tools/blob/main/src/examples/SPDXRdfExample-v2.0.rdf - raise issues like `'org.spdx.library.model.SpdxIdInUseException: Can not create Apache-2.0. It is already in use with type License which is incompatible with type ListedLicense')`. Tried with this 2.1 version as well but same issue - https://github.com/spdx/tools/blob/master/Examples/SPDXRdfExample-v2.1.rdf

The test files in the [tools-java](https://github.com/spdx/tools-java/tree/master/testResources) repo should work - as long as we're using the same version of the java tools.

BTW - I'll be offline for the next 3 days or so, so may be a bit slow to respond.

rtgdk:
@goneall Thanks for the help! The new files worked perfectly! Now all the tests in api are passing, except one which is getting skipped. Will try to run that too after setting up test repos for submit license tool.
Screenshot 2024-04-08 at 10 31 17 PM

goneall (approved):
Big improvement on the tests!  @rtgdk - if you're OK with fixing the remaining test issue in a separate PR, please go ahead and merge.

PR #538, opened 2024-08-01
add dots in iframe

ilans:
- Added a "Visual Editor" tab to the nav bar.
- Embedded "dots" in an iframe using https.
- Wrapped "block1" with a "dots" block so the iframe could stretch the entire height. If the "dots" block don's have any content then "block1" is rendered as usual, so it shouldn't affect existing templates.

### DIFF
--- a/src/app/templates/app/base.html
+++ b/src/app/templates/app/base.html
@@ -138,6 +138,7 @@ div p {
             <!-- <li><a href="/app/diff/" id="licensediffpage">License Diff</a></li> -->
             <li><a href="/app/xml_upload" id="license-xml-editor">License XML Editor</a></li>
             <li><a href="/app/ntia_checker/" id="checkerpage">NTIA Conformance Checker</a></li>
+            <li><a href="/app/dots" id="dots">Visual Editor</a></li>
             <li>
               <a class="dropdown-toggle" data-toggle="dropdown" href="#" id="licenserequest">License requests
                 <span class="caret"></span>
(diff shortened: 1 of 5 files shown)


src/app/templates/app/dots.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'app/base.html' %} {% load static %} {% block dots %}
+<iframe
+  src="http://151.145.81.233"
```
rtgdk, line 3: @ilans What feature does this dots iframe provides for the web tools?

ilans: Hi Rohit, "dots" is an open source visual SPDX tool for constructing and editing SPDXv3 documents. I created it as a playground for newcomers, but it could be useful for advanced users too. Give it a try...

I presented it at the technical and implementers meetings and was asked to add it to the tools website, which is what it was made for in the first place. The reason it's embedded in an iframe is that it's a React app (a pretty big one). It's also in active development and lives in its own repo: https://github.com/condots/dots

ilans: Note that I've changed the iframe src to embed dots using https.

ilans: Can I merge?

src/app/templates/app/base.html
```diff
@@ -181,13 +182,15 @@
     </div><!--/.nav-collapse -->
   </nav>
 
-    <div class="container body1">
```
rtgdk, line 184: Removing `container body1` div will cause changes on other pages.

We don't need to create separate block for dots. The frame will be visible below the spdx navigation bar.

src/app/templates/app/dots.html
rtgdk: @ilans @goneall Thanks for confirming. Left one more comment on the new commit to address before merge. Other changes lgtm.

src/app/templates/app/base.html
goneall: I'm not much of an HTML person, but I did go through all the pages with this pull request and didn't see any difference in the rendering.

Even though this change could impact other pages, it doesn't look like it actually does - so I'm OK merging as is.

@rtgdk - agree?

rtgdk: Yeah I agree. Merging this

PR #542, opened 2024-12-22
Add ntia-conformance-checker compliance selector to UI

bact:
Changes to ntia-conformance-checker related code, to prepare it for [more compliance checkers](https://github.com/spdx/ntia-conformance-checker/pull/224) in the future.

These changes are effectively only in the UI side for now and there's nothing change on the backend side apart from the upgrade of the library.

- Update ntia-conformance-checker version to 3.0.2
- Add compliance selector to UI
  - There's only one option now, which is the selected default "ntia"
- Add a (commented) code to get compliance request value
  - If the value is not exist, use "ntia" as default
  - This value will be used post-3.0.2 to let SbomChecker knows which compliance it should check against
- Remove unused imports
- Fix few typos

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,6 @@ gunicorn==22.0.0
 selenium==3.14.1
 django-oauth-toolkit==1.5.0
 django-rest-framework-social-oauth2==1.1.0
-spdx-tools==0.8.2
+spdx-tools==0.8.3
 ntia-conformance-checker==3.0.2
 -e git+https://github.com/spdx/spdx-license-matcher.git@v2.7#egg=spdx-license-matcher
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,5 +16,5 @@ selenium==3.14.1
 django-oauth-toolkit==1.5.0
 django-rest-framework-social-oauth2==1.1.0
 spdx-tools==0.8.2
-ntia-conformance-checker==1.1.0
+ntia-conformance-checker==3.0.2
 -e git+https://github.com/spdx/spdx-license-matcher.git@v2.7#egg=spdx-license-matcher
--- a/src/app/templates/app/ntia_conformance_checker.html
+++ b/src/app/templates/app/ntia_conformance_checker.html
@@ -18,7 +18,6 @@
       &nbsp; File Type &nbsp;
       <select name="format" id="format">
         <option value="0" data-value="0" selected="">-</option>
-        <option value="XLSX">XLSX Spreadsheet</option>
         <option value="XLS">XLS Spreadsheet</option>
         <option value="RDFXML">RDF/XML</option>
         <option value="TAG">Tag/Value</option>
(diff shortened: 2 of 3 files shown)


goneall (approved):
LGTM

requirements.txt
```diff
@@ -16,5 +16,5 @@ selenium==3.14.1
 django-oauth-toolkit==1.5.0
 django-rest-framework-social-oauth2==1.1.0
 spdx-tools==0.8.2
```
goneall, line 18:
```suggestion
spdx-tools==0.8.3
```

NTIA tools version 3.0.2 requires an update to the spdx-tools to 0.8.3

goneall (changes requested):
I'll do some more testing - may add further comments, but there is one change I found needs to be made with the requirements.txt file.

src/app/templates/app/ntia_conformance_checker.html
```diff
@@ -8,44 +8,49 @@
...
-		  <p class="file-name">No file selected</p>
-		</div>
-    </form>
+  <form id="checkform" enctype="multipart/form-data" class="form-horizontal" method="post">
+    {% csrf_token %}
+    <div class = "form-group">
+      &nbsp; File Type &nbsp;
+      <select name="format" id="format">
+        <option value="0" data-value="0" selected="">-</option>
+        <option value="XLSX">XLSX Spreadsheet</option>
```
goneall, line 21:
```suggestion
```

This option isn't supported by the Python libraries - we should remove it from the dropdown options.

src/app/templates/app/ntia_conformance_checker.html
```diff
@@ -8,44 +8,49 @@
...
+        <option value="XLS">XLS Spreadsheet</option>
```
goneall, line 22:
```suggestion
```

This option isn't supported by the Python libraries - we should remove it from the dropdown options.

src/app/templates/app/ntia_conformance_checker.html
```diff
@@ -8,44 +8,49 @@
...
+        <option value="RDFXML">RDF/XML</option>
+        <option value="TAG">Tag/Value</option>
+        <option value="JSON">JSON</option>
+        <option value="YAML">YAML</option>
```
goneall, line 26:
```suggestion
```

This option isn't supported by the Python libraries - we should remove it from the dropdown options.

goneall (changes requested):
A few more changes to fix problems that were present in the previous versions - might as well fix them with this update.

goneall (commented):
LGTM - It looks like we're about to get a new release of the conformance checker.

@bact - do you want to wait for the new release before merging?

goneall (approved):
Everything tests out fine, so I'm going to go ahead and merge this PR.

If we decide to include a new release of the NTIA Conformance Checker, we can create a second PR.

bact:
Thank you. I will create another PR after we have the new release of the checker.

PR #544, opened 2024-12-29
Descriptive title + Centred filename in drag&drop

bact:
- Use `<div>` (instead of `<p>`) for filename display in drag & drop area, so the text will be centred like other HTML elements in the same pane
- Update heading to match the tool title (see current Compare tool for an example)
- Put "V2" into File Type selections, for NTIA Conformance tool

--

**NTIA Conformance tool:** the heading is not descriptive and the filename ("No file selected") is off to the left

Screenshot 2024-12-29 at 14 51 47

**Compare tool:** the heading is descriptive and the filename is centred

Screenshot 2024-12-29 at 14 51 38


goneall (approved):
LGTM

Issue #527, opened 2024-01-21
check license button no longer works


goneall:
After updating to the latest release (1.2.3), clicking the check license button does not return anything.

In looking in the browser debug console message, it looks like it can't find the file `https://tools.spdx.org/static/js/utils.js`.

Perhaps the get static didn't execute successfully or it is in the wrong place.

Below is the console output:
```

﻿
submit_new_license/:422 
 GET https://tools.spdx.org/static/js/utils.js net::ERR_ABORTED 404 (Not Found)
submit_new_license/:896 Uncaught ReferenceError: findLicenseMatch is not defined
    at HTMLButtonElement.<anonymous> (submit_new_license/:896:5)
```

goneall:
On the production server, there is a static directory, but it doesn't contain the utils.js.

Here's the contents of `/spdx/src/app/static/js`:

```
bootstrap.js      editor          jquery-ui.min.js  npm.js
bootstrap.min.js  html2canvas.js  jquery.min.js     warning.js

```

goneall:
I could not locate the file `utils.js` in the docker instance.

I tried running the collect static command manually - here's the output:
[collect-static-output.txt](https://github.com/spdx/spdx-online-tools/files/14002636/collect-static-output.txt)

bact:
What is the expected behavior of the tool at https://tools.spdx.org/app/check_license/ ?

The interface itself does not very telling.

goneall:
> What is the expected behavior of the tool at https://tools.spdx.org/app/check_license/ ?
> 
> The interface itself does not very telling.

The expected behavior is to compare the submitted text against all listed licenses and listed license exceptions.  The result of the compare would be one of "Exact Match", "Close Match" or "No Match".  Exact match only if it matches using the license matching guidelines.  The implementation of close match is it achieve a high score on a text matching algorithm (e.g. Sørensen Dice) but does not pass the full matching guidelines.

There are two main use cases:
- Check a found license during an analysis of open source licenses for compliance reasons
- Check a license before submitting a new license to the license list - avoid duplication

bact:
Thank you

bact:
I have tried 

Apache 
Apache 2.0
Apache-2.0
GPL

Nothing matched

goneall:
> I have tried
> 
> Apache Apache 2.0 Apache-2.0 GPL
> 
> Nothing matched

It works from the full text of the license - not the license ID.

[...]

PR #545, opened 2024-12-30
Add usage description to Check License UI

bact:
- Add a usage description to Check License UI, based on https://github.com/spdx/spdx-online-tools/issues/527#issuecomment-2564809534
- Also add a missing closing `</div>` in the HTML


goneall (approved):
LGTM

PR #555, opened 2025-01-29
add missing iframe permissions

ilans:


ilans:
@goneall 

- allow-popups (required by dots): Allows opening new tabs or windows.
- allow-presentation (required by youtube): Allows to use external displays/screens (like Chrome Cast).

Neither of them allow the iframe to influence its parent app and they pose very low risks. Nevertheless, I spent a few hours trying to work around youtube's requirement so we could ditch the allow-presentation, but youtube won. We need both flags...

goneall:
Thanks @ilans for the additional research

I'll go ahead and merge

PR #559, opened 2025-03-06
Update generateXml.py

bhorsujal:
- Ensure `escapeXmlData` correctly assigns transformed string back.
- Format XML output with `toprettyxml(indent="  ")` for readability.
- Remove extra blank lines from the formatted XML output for clean structure.
- Add a check in `wrapBullets`.
- Simplify `groupLines` by replacing `math.floor()` with integer division.

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -106,7 +106,7 @@ def license_compare_helper(request):
                     compareclass.onlineFunction(callfunc)
                 except Exception as ex:
                     """Error raised by onlineFunction"""
-                    if (request.is_ajax()):
+                    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
                         ajaxdict["type"] = "warning2"
                         ajaxdict["files"] = filelist
                         ajaxdict["errors"] = errorlist
@@ -122,7 +122,7 @@ def license_compare_helper(request):
                     return result
                 if (warningoccurred==False):
                     """If no warning raised """
-                    if (request.is_ajax()):
+                    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
                         ajaxdict["medialink"] = settings.MEDIA_URL + folder + "/"+ rfilename
                         response = dumps(ajaxdict)
                         result['response'] = response
@@ -134,7 +134,7 @@ def license_compare_helper(request):
                     result['context'] = context_dict
                     return result
                 else :
-                    if (request.is_ajax()):
+                    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
                         ajaxdict["type"] = "warning"
                         ajaxdict["files"] = filelist
                         ajaxdict["errors"] = errorlist
@@ -151,7 +151,7 @@ def license_compare_helper(request):
                     result['context'] = context_dict
                     return result
             else :
-                if (request.is_ajax()):
+                if request.headers.get('x-requested-with') == 'XMLHttpRequest':
                     ajaxdict["files"] = filelist
                     ajaxdict["type"] = "error"
                     ajaxdict["errors"] = errorlist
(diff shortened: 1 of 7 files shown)


src/app/core.py
```diff
@@ -106,7 +106,7 @@ def license_compare_helper(request):
                     compareclass.onlineFunction(callfunc)
                 except Exception as ex:
                     """Error raised by onlineFunction"""
-                    if (request.is_ajax()):
+                    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
rtgdk, line 109: Create a util func for this and use that.

```
def is_ajax(request):
    return request.headers.get('x-requested-with') == 'XMLHttpRequest'
 ```
 
 Also, make sure this works instead of
 `request.META.get('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest'`

File: src/app/generateXml.py [hunk 2]
```diff
@@ -23,7 +23,7 @@ def previous_and_current(some_iterable):
 
 def escapeXmlData(string):
     for initial,final in list(entityMap.items()):
-        string.replace(initial, final)
+        string = string.replace(initial, final) # assign changed string back
```
rtgdk, line 26: Good catch

src/app/tests.py
```diff
@@ -648,7 +649,8 @@ class LicenseXMLEditorTestCase(StaticLiveServerTestCase):
     def setUp(self):
         options = Options()
         options.add_argument('-headless')
-        self.selenium = webdriver.Firefox(executable_path=GeckoDriverManager().install(), firefox_options=options)
+        service = Service(GeckoDriverManager().install())
```
rtgdk, line 652: Let's define this globally, instead of defining for each setUp

src/src/settings.py
```diff
@@ -234,7 +234,7 @@
 # URL Path Variables
 
 LOGIN_REDIRECT_URL = "/app/"
-REGISTER_REDIRECT_UTL = "/app/login/"
+REGISTER_REDIRECT_URL = "/app/login/"
```
rtgdk, line 237: Good catch

src/app/generateXml.py
```diff
@@ -154,4 +155,16 @@ def generateLicenseXml(licenseOsi, licenseIdentifier, licenseName, listVersionAd
...
+    # Format the XML string with indentation for better readability
+    xmlString = minidom.parseString(xmlString).toprettyxml(indent="  ")
+    
+    # Remove extra blank lines added by `toprettyxml()`, ensuring clean output
+    clean_lines = []
+    for line in xmlString.splitlines():
+        if line.strip():  # Only keep non-empty lines
+            clean_lines.append(line)
+
+    xmlString = "\n".join(clean_lines)
```
rtgdk, line 168: What does this code block do? Can you provide an example?

rtgdk (commented):
Left some comments

src/app/generateXml.py
bhorsujal: The code block formats an XML string to make it easier to read. First, it uses minidom.parseString().toprettyxml(indent="  ") to add proper indentation and newlines. Then it removes any extra blank lines that were added by toprettyxml(). This makes the final XML output clean and well-structured without affecting its functionality.

#### **Before Formatting (Raw XML)**
```xml
<root><child>Value</child><child2>Value1</child2></root>
```
#### After toprettyxml() (With extra blank lines)
```xml
<root>
  <child>Value</child>

  <child2>Value1</child2>

</root>
```
#### After Removing Extra Blank Lines (Final Output)
```xml
<root>
  <child>Value</child>
  <child2>Value1</child2>
</root>
```

src/app/core.py
bhorsujal: Sure, will do that!

src/app/generateXml.py
bhorsujal: Thanks!

src/src/settings.py
bhorsujal: Thanks!

src/app/tests.py
bhorsujal: Got that :+1:

bhorsujal:
Yes, I will work on the comments :+1: 
Thank you

bhorsujal:
I will raise a new PR and reference this conversation

PR #562, opened 2025-03-12
Restart SPDX on reboots and failures

goneall:
This change will restart SPDX if there is ever a reboot or failure.


rtgdk (approved):
LGTM

PR #564, opened 2025-03-28
Fix mismatch in error object structure in postscreenshot.js (function…

bhorsujal:
### Brief Explanation of Changes

1. Correctly matched the JSON structure returned by the Django view. The server returns the error type in `obj.data.type` and the message in `obj.message`.
2. Updated the `error:` callback in `postscreenshot.js` to check `if (obj.data && obj.data.type === "error")` and display `obj.message`, ensuring the modal shows the actual error text instead of `[object Object]`.

### DIFF
--- a/src/app/static/js/editor/postscreenshot.js
+++ b/src/app/static/js/editor/postscreenshot.js
@@ -54,12 +54,15 @@ async function postToGithub(data) {
             $("#modal-header").removeClass("green-modal");
             try {
             var obj = JSON.parse(e.responseText);
-            if (obj.type=="error"){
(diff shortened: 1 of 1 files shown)


rtgdk (approved):
Changes look good to me. @bhorsujal Can you add a test plan on how you tested this? Did you fake an error from server? Can you post the output/screenshot of the test?

bhorsujal:
Actually, I couldn't find a way to test it but by looking at the code I could clearly see the mismatch between the JSON Structure and callback. So that must be the issue here.

PR #569, opened 2025-07-30
Fix GitHub OAuth HTTPS redirect in production

### LINKED ISSUE #260: Redirect for the Github Authorization should use HTTPS rather than HTTP in production
I'm not sure if this is a security vulnerability, but the redirect probably should use HTTP rather than HTTPS.

Just changing it to HTTPS will likely mess up those running the spdx-online-tools in test environments or local development environments.

Perhaps the System configuration debug could be checked and the appropriate redirect made.

NOTE: the Github Auth configurations need to be updated to HTTPS as well as the code once this goes into production.

bhorsujal:
- Add SECURE_SSL_REDIRECT=True to force HTTPS in production
- Add SOCIAL_AUTH_REDIRECT_IS_HTTPS=True for OAuth redirects 
- Add USE_X_FORWARDED_HOST=True for proper proxy handling
- Settings only apply when DEBUG=False (production)

Fixes #260

### DIFF
--- a/src/src/settings.py
+++ b/src/src/settings.py
@@ -60,6 +60,10 @@ ALLOWED_HOSTS = ['*']
 
 if not DEBUG:
     ALLOWED_HOSTS = ['.tools.spdx.org', '52.32.53.255', '13.57.134.254']
+    # Force HTTPS redirects for OAuth in production
+    SECURE_SSL_REDIRECT = True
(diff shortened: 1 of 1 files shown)


bhorsujal:
It was not a critical bug fix. Just a security enhancement.
Kindly review.

BassCoder2808:
LGTM

BassCoder2808 (approved):

PR #573, opened 2025-09-07
Update Python version in Django CI and in Docker files

bact:
- Update Python version in Django CI to match with one in Docker file (3.10)
- Update Docker files' Python patch version to python:3.10.18 (fewer vulnerabilities)
- Remove deprecated "version" key in Docker compose files


goneall (approved):
LGTM

PR #596, opened 2025-11-01
conformance-checker: Use HTML output

### LINKED ISSUE #525: Use HTML format output for the NTIA conformance checker
as mentioned in #428 - there is a new HTML formatted output we can use for a prettier NTIA conformance checker output.

The changes would be in the app/core.py `ntia_check_helper` function.

bact:
Display output from `SbomChecker.output_html()` instead of the plain-text output.
To fix #525 

Examples:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ gevent==25.9.1
 gunicorn==23.0.0
 jpype1==1.6.0
 lxml>=5.4.0,<6.0.0
-ntia-conformance-checker==4.1.1
+ntia-conformance-checker==4.1.2
 python-dotenv>=1.0.1
 redis==4.4.4 # spdx-license-matcher depends on this specific version
 requests==2.32.5


src/app/core.py
```diff
@@ -199,48 +200,60 @@ def license_compare_helper(request):
...
             temp_stdout = StringIO()
             sys.stdout = temp_stdout
             schecker.print_components_missing_info()
-            schecker.print_table_output()
             sys.stdout = old_stdout
-            retval = temp_stdout.getvalue().replace(",",", ").replace("\n","<br/>")
+
+            html = schecker.output_html()
+            schecker_printout = re.sub(r"\n+", "\n", temp_stdout.getvalue().strip())
+            retval = schecker_printout.replace(",", ", ").replace("\n", "<br/>")
```
jspeed-meyers: This looks good to me. I feel bad that there needs to be convoluted regular expressions after `schecker.output_html()`. But, like you alluded in the PR, perhaps this could be cleaned up in a future version?

jspeed-meyers:
@bact: I left one minor comment. Like usual, nice work.

src/app/core.py
bact: That line is because we still need information about missing components and it is not available yet in HTML output. So we still have to rely on the output of print_components_missing_info in stdout, and thus the regex treatment.

Yes. I will looking at how to put the missing component info into HTML output in the next version of ntia-conformance-checker.

bact: Open the issue in ntia-conformance-checker as https://github.com/spdx/ntia-conformance-checker/issues/320

bact: Thank you too for the review :) 
I will put this as resolved for now and we will follow up in ntia-conformance-checker repo.

bact:
Wait for one fix in https://github.com/spdx/ntia-conformance-checker/pull/326

PR #616, opened 2025-12-26
Report License Matcher and Python Model versions

bact:
Report versions of these packages in the About page, to assist debugging and also make awareness to the packages:

- SPDX License Matcher
  - was omitted, put back after https://github.com/spdx/spdx-license-matcher/pull/30 got merged and has new release (2.8) https://github.com/spdx/spdx-license-matcher/releases/tag/v2.8
- SPDX Python Model

Also remove unused `os` import in src/app/views.py

PR #618, opened 2025-12-27
Modify Docker build command to use 'no-cache' option

goneall:
Updated deployment instructions to include 'no-cache' option for building Docker images.

The docker image may not get properly rebuilt if caching is enabled and only some of the source files have changed.


bact:
See https://github.com/spdx/spdx-online-tools/pull/616#issuecomment-3694136863 for context

bact (approved):

PR #626, opened 2026-04-07
Move media cleanup scheduling out of the web app container

### LINKED ISSUE #625: Disk space cleanup not being run
It looks like we are no longer running the disk space cleanup utility introduced with PR #440 

The [changes to dockerfile.prod](https://github.com/spdx/spdx-online-tools/pull/440/changes#diff-59773410e08ccbb17158a8611e631a128e99a191864c04d194cadcf973def4f4) seems to have been removed at some point.

I looked back through git blame and could not find where this occurred.

@rtgdk @BassCoder2808 @bact let me know if you recall if this was intentional.  If not, perhaps we can just add it back.

bact:
To bring back the media cleanup task, which was dropped at some point during recent updates, as reported by #625.

Previously, media cleanup was scheduled inside the web app container. That setup can trigger duplicate cleanup runs when multiple web app containers are running.

This change moves scheduling to host cron and refactors cleanup into a Django management command. The command can now be executed consistently from host cron and from Docker CLI commands.

Tested with `python src/manage.py test app.tests.TestCronJob`

### DIFF
--- a/crontab
+++ b/crontab
@@ -4,4 +4,5 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 # Every day at 15:30 UTC, finds the container by docker-compose
 # service label ("spdx"), and runs the media cleanup job.
-30 15 * * * c=$(docker ps -qf "label=com.docker.compose.service=spdx" | head -n 1) && [ -n "$c" ] && docker exec "$c" /usr/local/bin/python /spdx/src/manage.py cleanup_media
+# If no such container is running, logs a warning to the same log file used by the cleanup script.
+30 15 * * * c=$(docker ps -qf "label=com.docker.compose.service=spdx" | head -n 1) && [ -n "$c" ] && docker exec "$c" /usr/local/bin/python /spdx/src/manage.py cleanup_media || { d=$(docker inspect -f '{{index .Config.Labels "com.docker.compose.project.working_dir"}}' $(docker ps -aqf "label=com.docker.compose.service=spdx" | head -n 1) 2>/dev/null); echo "$(date -u +'\%Y-\%m-\%d \%H:\%M:\%S,000') - app.cleanup - WARNING - cleanup_media: no running container" >> "$d/container_logs/deletedFiles.log"; }
--- a/crontab
+++ b/crontab
@@ -1,2 +1,6 @@
-# Run the cleanup script every day at 3:30 pm
-30 15 * * * python /spdx/app/scripts/cleanup.py
+# Host cron template (install on the EC2 host, not inside Docker).
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# Run the Django media cleanup job every day at 15:30 UTC.
+30 15 * * * sh -c 'if docker ps --format "{{.Names}}" | grep -qx spdx_prod; then docker exec spdx_prod /usr/local/bin/python /spdx/src/manage.py cleanup_media; else echo "spdx_cleanup skipped: container spdx_prod is not running"; fi' 2>&1 | logger -t spdx_cleanup
--- a/crontab
+++ b/crontab
@@ -3,4 +3,4 @@ SHELL=/bin/sh
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 # Run the Django media cleanup job every day at 15:30 UTC.
-30 15 * * * sh -c 'if docker ps --format "{{.Names}}" | grep -qx spdx_prod; then docker exec spdx_prod /usr/local/bin/python /spdx/src/manage.py cleanup_media; else echo "spdx_cleanup skipped: container spdx_prod is not running"; fi' 2>&1 | logger -t spdx_cleanup
+30 15 * * * docker exec spdx_prod /usr/local/bin/python /spdx/src/manage.py cleanup_media
--- a/crontab
+++ b/crontab
@@ -2,5 +2,6 @@
 SHELL=/bin/sh
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-# Run the Django media cleanup job every day at 15:30 UTC.
-30 15 * * * docker exec spdx_prod /usr/local/bin/python /spdx/src/manage.py cleanup_media
+# Every day at 15:30 UTC, finds the container by docker-compose
+# service label ("spdx"), and runs the media cleanup job.
+30 15 * * * c=$(docker ps -qf "label=com.docker.compose.service=spdx" | head -n 1) && [ -n "$c" ] && docker exec "$c" /usr/local/bin/python /spdx/src/manage.py cleanup_media
--- a/src/app/management/commands/cleanup_media.py
+++ b/src/app/management/commands/cleanup_media.py
@@ -1,5 +1,4 @@
 import argparse
-from datetime import datetime, timezone
 
 from django.core.management.base import BaseCommand
 
@@ -25,17 +24,4 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
-        started_at = datetime.now(tz=timezone.utc).isoformat()
-        self.stdout.write(f"Cleanup started at {started_at}")
-
-        deleted_files = clean_media(days_threshold=options["days_threshold"])
-        for file_info in deleted_files:
-            self.stdout.write(
-                f"Deleted file: {file_info['name']} (modified at {file_info['modified_at']})"
-            )
-
-        self.stdout.write(
-            self.style.SUCCESS(
-                f"Cleanup completed; deleted {len(deleted_files)} file(s)."
-            )
-        )
+        clean_media(days_threshold=options["days_threshold"])
--- a/src/app/management/commands/cleanup_media.py
+++ b/src/app/management/commands/cleanup_media.py
@@ -29,4 +29,5 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
-        clean_media(days_threshold=options["days_threshold"])
+        deleted = clean_media(days_threshold=options["days_threshold"])
+        self.stdout.write(f"Deleted {len(deleted)} file(s).")
--- a/src/app/management/commands/cleanup_media.py
+++ b/src/app/management/commands/cleanup_media.py
@@ -3,7 +3,7 @@
 
 """Django management command for cleaning up old media files."""
 
-import argparse
+from argparse import ArgumentTypeError
 
 from django.core.management.base import BaseCommand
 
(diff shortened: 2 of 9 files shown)


goneall:
I think this approach should work.  I can test this on the production server after it is merged.

goneall (approved):
I didn't see any problems on review - although I'm not much of a Python expert.  @rtgdk if you get a chance - pls add another review.

crontab
```diff
@@ -1,2 +1,6 @@
-# Run the cleanup script every day at 3:30 pm
-30 15 * * * python /spdx/app/scripts/cleanup.py
+# Host cron template (install on the EC2 host, not inside Docker).
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# Run the Django media cleanup job every day at 15:30 UTC.
+30 15 * * * docker exec spdx_prod /usr/local/bin/python /spdx/src/manage.py cleanup_media
```
rtgdk: The container name `spdx_prod` is hardcoded. If the production docker-compose project name or service name changes, this silently breaks. Consider using `docker compose exec <service>` instead, or at minimum document this assumption.

src/app/scripts/cleanup.py
```diff
@@ -1,32 +1,61 @@
...
+from pathlib import Path
+from time import time
+
+from django.conf import settings
+
+ANONYMOUS_MEDIA_SUBDIR = "AnonymousUser"
+DEFAULT_DAYS_THRESHOLD = 10
+SECONDS_PER_DAY = int(timedelta(days=1).total_seconds())
+LOG_FILE = Path(settings.PROJECT_ROOT) / "container_logs" / "deletedFiles.log"
+LOG_FILE.parent.mkdir(parents=True, exist_ok=True)
```
rtgdk: `mkdir` runs as a side-effect when the module is imported (including during tests, management command discovery, etc.). This could create unexpected directories. Consider moving the mkdir into the clean_media() function or into the logger setup block.

src/app/scripts/cleanup.py
```diff
@@ -1,32 +1,61 @@
...
 
-logger.addHandler(handler)
+    if not media_dir.is_dir():
+        logger.info("Cleanup skipped because directory does not exist: %s", media_dir)
+        return deleted_files
 
-def cleanMedia():
-    MEDIA_DIR = os.path.join(settings.MEDIA_ROOT, "AnonymousUser")
-    DAYS_THRESHOLD = 10
+    for filepath in sorted(media_dir.iterdir()):
```
rtgdk, line 57: nit: Why do we need to sort the media directory files?

rtgdk (commented):
Overall looks good to me. Left some comments to address

src/app/scripts/cleanup.py
bact: This is to have the filenames of deleted files sorted in the log.

If not desired, can remove the sort.

rtgdk: Okay, let's keep it.

crontab
bact: Updated to dynamically find the container by service name.

src/app/scripts/cleanup.py
bact: Updated to put mkdir in a logger setup block.

src/app/scripts/cleanup.py
```diff
@@ -1,32 +1,70 @@
...
+            continue
 
-    now = time.time()
+        modified_at = filepath.stat().st_mtime
+        if (now - modified_at) <= cutoff_seconds:
+            continue
 
-    # log the time of the cron job
-    logger.info('Cron job ran at %s', now)
+        filepath.unlink()
```
rtgdk: Two issues here:

1. The log says "Deleting" but fires *after* `unlink()` — the file is already gone. If `unlink()` raises, no log is emitted at all.
2. If one file fails to delete (e.g. permission error), the entire cleanup aborts — remaining files are skipped.

Consider wrapping in a try/except and changing the verb to "Deleted":

```suggestion
        try:
            filepath.unlink()
        except OSError:
            logger.exception("Failed to delete %s", filepath.name)
            continue
        logger.info("Deleted file %s with date %s", filepath.name, datetime.fromtimestamp(modified_at, tz=timezone.utc).isoformat())
```

src/app/management/commands/cleanup_media.py
```diff
@@ -0,0 +1,32 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--days-threshold",
+            type=_non_negative_int,
+            default=10,
+            help="Delete files older than this many days.",
+        )
+
+    def handle(self, *args, **options):
+        clean_media(days_threshold=options["days_threshold"])
```
rtgdk: `handle()` calls `clean_media()` but never writes to `self.stdout`. When an operator runs `manage.py cleanup_media` manually, there's no visible feedback — they'd have to check the log file to know anything happened. Consider adding a summary line:

```suggestion
        deleted = clean_media(days_threshold=options["days_threshold"])
        self.stdout.write(f"Deleted {len(deleted)} file(s).")
```

src/app/scripts/cleanup.py
```diff
@@ -1,32 +1,70 @@
...
 
-formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
-handler.setFormatter(formatter)
 
-logger.addHandler(handler)
+def _setup_logger():
+    log_file = Path(settings.PROJECT_ROOT) / "container_logs" / "deletedFiles.log"
+    
+    if not any(
+        isinstance(handler, logging.FileHandler) and Path(handler.baseFilename) == log_file
```
rtgdk: `handler.baseFilename` is a `str` while `log_file` is a `Path`. The `==` comparison works on most platforms, but can give false negatives on case-insensitive filesystems or when symlinks are involved. Safer to resolve both sides:

```suggestion
        isinstance(handler, logging.FileHandler) and Path(handler.baseFilename).resolve() == log_file.resolve()
```

crontab
```diff
@@ -1,2 +1,7 @@
-# Run the cleanup script every day at 3:30 pm
-30 15 * * * python /spdx/app/scripts/cleanup.py
+# Host cron template (install on the EC2 host, not inside Docker).
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# Every day at 15:30 UTC, finds the container by docker-compose
+# service label ("spdx"), and runs the media cleanup job.
+30 15 * * * c=$(docker ps -qf "label=com.docker.compose.service=spdx" | head -n 1) && [ -n "$c" ] && docker exec "$c" /usr/local/bin/python /spdx/src/manage.py cleanup_media
```
rtgdk: If the container is down, `[ -n "$c" ]` fails and the job silently exits with success. For a daily job this is low-risk, but a prolonged outage (container down for weeks) would go unnoticed. Consider logging on failure so operators can spot it. Adjust the log path to match your deployment:

```suggestion
30 15 * * * c=$(docker ps -qf "label=com.docker.compose.service=spdx" | head -n 1) && [ -n "$c" ] && docker exec "$c" /usr/local/bin/python /spdx/src/manage.py cleanup_media || echo "$(date -u +\%FT\%TZ) cleanup_media: no running spdx container found" >> /path/to/container_logs/deletedFiles.log
```

src/app/tests.py
```diff
@@ -1592,9 +1593,8 @@ def test_no_license_id_on_license_requests_edit_xml(self):
         self.assertEqual(resp.resolver_match.func.__name__,"licenseNamespaceRequests")
 
 class TestCronJob(TestCase):
-    def test_delete_old_files(self):
-        """Check if the files older than 10 days are getting deleted or not"""
-        # create a test directory with some files
+    def test_clean_media_deletes_only_expired_files(self):
+        """ Check if the files older than 10 days are getting deleted"""
         test_dir = os.path.join(settings.MEDIA_ROOT, 'AnonymousUser')
```
rtgdk, line 1599: The test creates files under the real `settings.MEDIA_ROOT`. If the test fails mid-run (e.g. an assertion before the cleanup loop), those files are left behind. Consider registering a cleanup callback, or using `tempfile.mkdtemp()` with `clean_media(media_root=...)` since that parameter already exists.

(Would need `import shutil` at the top of the file.)

```suggestion
        test_dir = os.path.join(settings.MEDIA_ROOT, 'AnonymousUser')
        os.makedirs(test_dir, exist_ok=True)
        self.addCleanup(shutil.rmtree, test_dir, ignore_errors=True)
```

src/app/scripts/cleanup.py
```diff
@@ -1,32 +1,70 @@
...
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from time import time
+
+from django.conf import settings
 
-logger = logging.getLogger('my_logger')
+ANONYMOUS_MEDIA_SUBDIR = "AnonymousUser"
+DEFAULT_DAYS_THRESHOLD = 10
+SECONDS_PER_DAY = int(timedelta(days=1).total_seconds())
```
rtgdk: Nit: 86400 is the universal constant for seconds-in-a-day. The `timedelta` indirection computes the same value but doesn't add clarity — it actually makes the reader pause to verify it's doing what they expect.

```suggestion
SECONDS_PER_DAY = 86400
```

src/app/management/commands/cleanup_media.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: 2026-present SPDX contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Django management command for cleaning up old media files."""
+
+import argparse
```
rtgdk: Nit: `argparse` is only imported for `ArgumentTypeError`. A targeted import is slightly clearer about intent, and the reference on line 16 becomes just `raise ArgumentTypeError(...)`.

```suggestion
from argparse import ArgumentTypeError
```

src/app/scripts/cleanup.py
bact: Thank you. Done.

rtgdk (approved):

src/app/management/commands/cleanup_media.py
bact: Print status to stdout now.

crontab
bact: Thank you. It is now do the log when container is not available.

src/app/tests.py
bact: Done.

bact:
Fix all the issues raised by @rtgdk. Thank you for the reviews and suggestions.

bact:
With two approvals, I will merge this - so we can manage any merge conflicts with the tests at #628.

The crontab setup at host level is still needed.

Will open another PR to fix this if it fails in production environment.

rjb4standards:
The online validation tool is functioning properly again. Thanks to the team for fixing the problem.

goneall:
Thanks @bact and @rtgdk

PR #627, opened 2026-04-08
Add Postgres to docker-compose.prod.yml

### LINKED ISSUE #617: Move the production PostgreSQL DB to the local machine
We're currently using an AWS RDB installation which is rather expensive and overkill for how it is being used.

To migrate the database, we'll need to:

- [ ] Update the production docker-compose.prod.yml file to create the PostgreSQL container
- [ ] Change the DB configuration for the Django app to point to the local PostgreSQL DB
- [ ] Migrate the data from the RDB database to the local database
- [ ] Shut down the RDB database

goneall:
Fixes #617

Note that the production environment needs to be migrated.

This involves:
1. Build with the updated docker-compose.prod.yml file
2. Update the spdx-prod.env file per the updated documentation
3. Run `docker-compose -f docker-compose.prod.yml up -d`
4. watch for any errors, especially with the postgres container initialization
5. Stop the SPDX online tools app
6. backup the RDS database
7. Restore the RDS database to the local database
8. Start the SPDX online tools app

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -25,6 +25,17 @@ services:
     entrypoint: "/bin/sh -c 'trap exit TERM; while :; do certbot renew; sleep 12h & wait $${!}; done;'"
     restart: always
 
+  postgres:
+    container_name: container-pg
+    image: postgres
+    hostname: localhost
+    ports:
+      - "5432:5432"
+    env_file: spdx-prod.env
+    volumes:
+      - postgres-data:/var/lib/postgresql
+    restart: unless-stopped
+
   spdx:
     container_name: spdx_prod
     build:
@@ -61,3 +72,4 @@ volumes:
   data-volume:
   static_volume:
   media_volume:
+  postgres-data:
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -28,7 +28,6 @@ services:
   postgres:
     container_name: container-pg
     image: postgres:18-alpine
-    hostname: localhost
     ports:
       - "5432:5432"
     env_file: spdx-prod.env
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -56,7 +56,7 @@ services:
       - SPDXParser.LocalLicensesDir=/licenses/current
       - SPDXParser.OnlyUseLocalLicenses=true
     depends_on:
-      - redis
+      - postgres
     restart: always
 
   redis:
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -57,6 +57,7 @@ services:
       - SPDXParser.OnlyUseLocalLicenses=true
     depends_on:
       - postgres
+      - redis
     restart: always
 
   redis:
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -27,7 +27,7 @@ services:
 
   postgres:
     container_name: container-pg
-    image: postgres
+    image: postgres:18-alpine
     hostname: localhost
     ports:
       - "5432:5432"
--- a/README-PRODUCTION.md
+++ b/README-PRODUCTION.md
@@ -81,12 +81,6 @@ Following are the steps for a clean initial installation of the application:
   - Create an IAM role using the EC2 use case with the policy AmazonEC2ContainerRegistryPowerUser
   - Attach the role to the EC2 instance
 - Create a repository in the Amazon ECR spdx/online-tools
-- Create an RDS database
-  - Instance name `spdx-online-tools-db-production`
-  - Username `spdx`
-  - Set a password which will be used later in setting some of the environment variables in docker
-  - Under connectivity drop down, add VPC security group `spdx-online-tools-ec2`
-  - Change initial database name to `spdx_db` (under additional config)
 - Add an inbound rule to for PostgreSQL access
   - Edit the security group inbound rules for spdx-online-tools-ec2
   - Add PostgreSQL with a source of the EC2 instance (search for spdx-)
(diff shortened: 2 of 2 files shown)


bact:
>     6. backup the RDS database
> 
>     7. Restore the RDS database to the local database

Due to https://github.com/spdx/spdx-online-tools/issues/629#issuecomment-4226402943, we may like to rebuild the database.

rtgdk (commented):
The migration away from RDS makes sense for simplifying the deployment. A few concerns with the current configuration:

1. **`SQL_HOST=localhost` won't work from the spdx container.** In Docker's default bridge network, `localhost` inside the spdx container refers to itself, not the postgres container. It should be `SQL_HOST=postgres` (the Compose service name) or `SQL_HOST=container-pg` (the container name). The `hostname: localhost` directive only sets the *internal* hostname of the postgres container -- it doesn't affect DNS resolution from other containers.

2. **Missing `depends_on`**: The `spdx` service depends on `redis` but not on `postgres`. Adding `postgres` to `depends_on` would ensure correct startup order.

3. **Volume path**: The standard Postgres data directory is `/var/lib/postgresql/data`, not `/var/lib/postgresql`. Using the parent directory may cause permission issues or data not persisting correctly. Consider `postgres-data:/var/lib/postgresql/data`.

4. **Unpinned image**: `image: postgres` will pull `latest`, which could break on a major Postgres version change. Consider pinning to a specific version (e.g., `postgres:16-alpine`).

5. **`POSTGRES_DB=postgresql`** vs **`SQL_DATABASE=spdx_db`**: These are different values. `POSTGRES_DB` controls what database the postgres container creates on first init. For the spdx app to connect successfully, `POSTGRES_DB` should match `SQL_DATABASE` (i.e., both should be `spdx_db`).

docker-compose.prod.yml
bact, line 59:
```suggestion
      - postgres
```

bact (changes requested):

docker-compose.prod.yml
```diff
@@ -25,6 +25,17 @@ services:
     entrypoint: "/bin/sh -c 'trap exit TERM; while :; do certbot renew; sleep 12h & wait $${!}; done;'"
     restart: always
 
+  postgres:
+    container_name: container-pg
+    image: postgres
```
bact:
```suggestion
    image: postgres:18-alpine
```

Pin major version.

Per @rtgdk suggestions

bact (changes requested):

docker-compose.prod.yml
```diff
@@ -25,6 +25,17 @@ services:
...
+    hostname: localhost
+    ports:
+      - "5432:5432"
+    env_file: spdx-prod.env
+    volumes:
+      - postgres-data:/var/lib/postgresql
```
bact, line 35:
```suggestion
      - postgres-data:/var/lib/postgresql/data
```

bact (changes requested):

goneall:
Thanks @rtgdk!  You probably saved me several hours of debugging!

for 3, the choice was of the volume path was intentional.  I initially had a volume path of `/var/lib/postgresql/data` and I got an error while testing.  Turns out they recently changed where the volume should point with Postgres 18 (see [this reddit comment](https://www.reddit.com/r/selfhosted/comments/1nsjxm2/if_youre_moving_to_docker_postgres_18_you_should/)).

The other changes look like important improvements.

docker-compose.prod.yml
goneall: We should leave this as is since we are using postgres version 18.

goneall:
> > ```
> > 6. backup the RDS database
> > 
> > 7. Restore the RDS database to the local database
> > ```
> 
> Due to [#629 (comment)](https://github.com/spdx/spdx-online-tools/issues/629#issuecomment-4226402943), we may like to rebuild the database.

@bact - I think #629 refers to the redis database, not the postgresql database.

goneall:
for 5 above, I'll need to restore the database from the backup.  It's only a bit more work to create the SPDX database first.  I left the default postgres database since that was how the current database is setup.  That being said, I can't think of any reason why we need these to be different.

docker-compose.prod.yml
```diff
@@ -45,7 +56,7 @@ services:
       - SPDXParser.LocalLicensesDir=/licenses/current
       - SPDXParser.OnlyUseLocalLicenses=true
     depends_on:
-      - redis
+      - postgres
```
goneall, line 58: Looks like we removed the redis dependency, I'll add it back

goneall, line 58:
```suggestion
      - postgres
      - redis
```

goneall (commented):
Add back redis to SPDX dependencies

bact (approved):

goneall:
@rtgdk for number 1 above - from what I read, this line isn't really needed in the docker-compose file - it defaults to naming the hostnames in a way that makes the services accessible between the containers.

I think I should just get rid of that line.  Let me know if you agree.

rtgdk (commented):
Thanks for addressing the image pinning and the volume path explanation (TIL about the Postgres 18 change). A few items that still look unresolved in the current diff:

1. **`depends_on` is missing `redis`**: The current diff replaces `- redis` with `- postgres`, but the spdx container still uses `SPDX_REDIS_HOST=redis`. I see @goneall posted an inline suggestion to add both, but it doesn't appear to be committed yet.

2. **`hostname: localhost`**: @goneall mentioned removing this line -- agreed, it's misleading since it doesn't affect inter-container DNS resolution.

3. **`SQL_HOST=localhost` in the docs**: This is the one I'm most concerned about. In Docker's default bridge network, `localhost` inside the spdx container refers to itself. The documented `SQL_HOST` should be `postgres` (the Compose service name) for the spdx container to reach the database.

goneall:
Thanks @rtgdk for the additional review.

I think everything should be resolved:

> Thanks for addressing the image pinning and the volume path explanation (TIL about the Postgres 18 change). A few items that still look unresolved in the current diff:
> 
> 1. **`depends_on` is missing `redis`**: The current diff replaces `- redis` with `- postgres`, but the spdx container still uses `SPDX_REDIS_HOST=redis`. I see @goneall posted an inline suggestion to add both, but it doesn't appear to be committed yet.

I committed the suggestion

> 2. **`hostname: localhost`**: @goneall mentioned removing this line -- agreed, it's misleading since it doesn't affect inter-container DNS resolution.

This is now removed

> 3. **`SQL_HOST=localhost` in the docs**: This is the one I'm most concerned about. In Docker's default bridge network, `localhost` inside the spdx container refers to itself. The documented `SQL_HOST` should be `postgres` (the Compose service name) for the spdx container to reach the database.

I updated the README to use the correct host name.

PR #628, opened 2026-04-09
Fix tests

### LINKED ISSUE #530: Fix all api and app Unit Tests to run successfully on all environments

bact:
Fix test CI, update the branch name from `master` to `main`, so it's running again.

Fix test cases, most of them involves three categories:

1) Updating path resolution, due to changing directory structure
    - Convert `"examples/file.xls"` --> `getExamplePath("file.xls")`

2) Add few delay before doing UI actions + assertion
    - Add `WebDriverWait(driver, 10).until(...)`

3) Cover minor browsers differences 
    - Convert `driver.find_element(By.ID, "button-id").click()` --> `driver.execute_script("arguments[0].click();", driver.find_element(By.ID, "button-id"))`

--

Also fix the formatxml.py call. It was broken earlier due to wrong path resolution.
There are 6 tests that rely on the formatted XML content. So need this in order to get all the tests working.


bact:
Cache for geckodriver is added to the workflow step, since some times 10 tests will be skipped because the geckodriver is not available. (see [this run](https://github.com/spdx/spdx-online-tools/actions/runs/24279257883/job/70898403549#step:9:558), for example)

geckodriver is downloaded from Mozilla GitHub repo over the network, so it may unreliable or it may reached rate limit if we run the tests a lot.

The [remaining 6 tests](https://github.com/spdx/spdx-online-tools/actions/runs/24281579625/job/70904197091#step:11:540) that still skipping are tests that required GitHub login, like the license request task.

With these tests and the test workflow runing again, we should have more confidence adding/modifying things looking forward.

goneall (approved):
Thanks @bact 

Changes look good to me

@rtgdk - if you get a chance, can you do a review?

rtgdk (approved):
Looks good!

PR #635, opened 2026-04-11
Update Redis commands in README

bact:
The suggested command line does not work on Apple Silicon.

Suggests brew command that works for both Apple Silicon and Intel Macs instead.


goneall (approved):
LGTM

PR #636, opened 2026-04-13
Remove unused code

bact:
Fixes by `ruff check --fix`


goneall (commented):
Looks OK to a non-Python programmer ;)  - @rtgdk - could you have a quick look?

rtgdk (approved):
LGTM. The `except UserSocialAuth.DoesNotExist as AttributeError` fix is a genuine bug fix, not just a style change -- it was silently shadowing the built-in `AttributeError` name. Good catch by ruff.

bact:
Thank you.

PR #637, opened 2026-04-24
Update JPype to 1.7.0

### LINKED ISSUE #634: Update JPype to 1.7.0
A lot of fixes, some of them are about race condition.
https://github.com/jpype-project/jpype/releases/tag/v1.7.0
https://pypi.org/project/jpype1/

Should check if it is compatible with our code base that use 1.6.0 and upgrade.

Maybe after the updated tests in #628 got merged.

bact:
- Update JPype to 1.7.0
- Use context manager for JVM attach/detach, to ensure that thread will always get detached.

--

- [x] Works locally with validate, convert, compare (tools-java)
- [x] Works locally with license check (spdx-license-matcher)

To resolve #634

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -5,12 +5,14 @@
 """This file contains the core logic used in the SPDX Online Tools' APP and API"""
 
 import os
+from contextlib import contextmanager
 from json import dumps
 from time import time
 from traceback import format_exc
 from urllib.parse import urljoin
 
 import jpype
+import jpype.imports  # noqa: F401 - required for Java package imports
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
 from django.utils.datastructures import MultiValueDictKeyError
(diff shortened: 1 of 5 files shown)


rtgdk (commented):
Strong improvement to JVM lifecycle management. The context manager approach is much safer than the manual attach/detach calls scattered across views. A few observations:

1. The PR checklist shows "Works locally with license check (spdx-license-matcher)" is still unchecked. Is that blocking, or is it expected to work and just not yet verified?

2. `_jvm_thread()` is prefixed with underscore (private convention) but is used across modules (`api/views.py`, `app/views.py`). Consider naming it `jvm_thread()` to signal it's part of the module's public API.

3. The `isAttached()` guard in the finally block is a nice touch for the spdx_license_matcher v2.8 double-detach issue. Worth noting whether this workaround can be removed if/when spdx_license_matcher is updated.

bact:
To make sure there will be no conflict in the thread management style, will wait until spdx-license-matcher upgrade to JPype 1.7.0 and use similar attach/detach style.

An PR is open there:

- https://github.com/spdx/spdx-license-matcher/pull/43

src/app/core.py
```diff
@@ -21,33 +23,47 @@
...
         )
-    # Attach a thread to JVM and start processing
-    jpype.attachThreadToJVM()
-    jpype.JPackage("org.spdx.library").SpdxModelFactory.init()
+        from org.spdx.library import SpdxModelFactory
+        SpdxModelFactory.init()
+
+
+@contextmanager
+def _jvm_thread():
```
goneall: Should we rename to `jvm_thread` as suggested by @rtgdk [here](https://github.com/spdx/spdx-online-tools/pull/637#pullrequestreview-4195375350)?

goneall (changes requested):
Overall looks fine - just one minor suggestion

src/app/core.py
bact: i will update that

bact: fixed.

goneall (approved):

goneall:
I can create a release of the spdx-license-matcher once the PR's are merged.

rtgdk (approved):
Thanks for addressing the naming feedback -- `jvm_thread()` reads much better as a public API. The upstream coordination with spdx-license-matcher makes sense. LGTM.

goneall:
I just released a new version of the license matcher (v2.9) with the updated JPype implementation.

I created PR #649 to update the license matcher to the latest version.

bact:
The failed test (with `TimeoutException`) is Selenium browser test which is not related to the JPype (it is the test of XML editor split view, which does not use Java). I will try to increase the wait time for browser DOM update, for test stability.

bact:
Manually tested locally. Each function works as expected.

Rerun the test on GitHub and it passes the test now.

PR #642, opened 2026-04-25
Front page redesign - big buttons

bact:
- Make the front page more functional by providing buttons to apps immediately
- Buttons will flow and stack by the screen width
- Redesign the top pane, order the tab logically by category (SBOM tools, license tools, about)
- Make sure all the apps (in each tab) are being displayed in a consistent way

On wider screen:

On narrower screen:

### DIFF
--- a/src/app/static/css/dragdrop.css
+++ b/src/app/static/css/dragdrop.css
@@ -1,48 +1,57 @@
 #drop-area {
-    border: 2px dashed rgb(221, 221, 221);
-    border-radius: 20px;
-    width: 90%;
-    min-height: 220px;
-    font-family: sans-serif;
-    margin: 30px auto;
-    padding: 20px 0;
-    transition: box-shadow 0.2s;
-    overflow: hidden;
+    border: 2px dashed #a8c8d8;
+    border-radius: 10px;
+    width: 100%;
+    max-width: 600px;
+    min-height: 150px;
+    margin: 16px auto;
+    padding: 24px 20px 16px;
+    text-align: center;
+    background: #f4fafd;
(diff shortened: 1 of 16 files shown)


goneall:
I like the new look.  As long as the license section doesn't scroll off the bottom, this should work.

src/app/static/css/dragdrop.css
goneall, line 1: Since we're changing a style sheet currently in use, have we checked to make sure it doesn't adversely affect any other web pages?  I could do a quick search to verify - but I didn't want to duplicate effort.

bact:
The 'hamburger' menu in the narrow screen mode can scroll and expand, all license options are accessible:

goneall (commented):
I reviewed the files and didn't see any issues - just one question on verifying the modification common CSS files didn't impact HTML files not reviewed / included in this PR.

src/app/static/css/dragdrop.css
bact: Thanks for flagging. They are all already checked. All of them functions as expected.

bact: I will go another round of testing. Just to be sure.

bact: I have just adjusted the vertical spacing of the License XML Editor. While not really effecting funcationality, the new top menu bar slightly overlapped the top tabs of the License XML Editor. Fixed now. No longer overlapped. Editor tabs are fully displayed.

goneall (approved):
LGTM

bact:
I will merge this so we can test it early with other changes as well.

PR #644, opened 2026-04-26
Generate web API doc

### LINKED ISSUE #91: Add support for documenting API

bact:
Use [drf-spectacular](https://github.com/tfranzel/drf-spectacular) to generate API doc and schema of Django REST API.

The API doc will be available at https://tools.spdx.org/api/ (same namespace with endpoints).
Once it is up online, we can put the link in About page.

To fix #91

- `/auth/` endpoints are not included since they are 3rd-party and already listed at https://tools.spdx.org/auth/convert-token/
- `/api2/` endpoints are also included, and their APIs are already listed at https://tools.spdx.org/api2/

The `/api/` endpoints API doc will look like this:

With the generated schema, we can also use it for Redoc to generate another sytle of API doc as well:

Screenshot 2026-04-26 at 17 48 14


bact:
Btw, the endpoints listed in https://tools.spdx.org/api2/ looks obsoleted 

The comment from https://github.com/spdx/spdx-online-tools/commit/687b76f4867c83d0f5216ec44724ba415c32b307 said the endpoints are moved from `/api2/` to `/api/`.

rtgdk (approved):
Great addition -- the API docs look polished and the schema annotations are thorough.

A couple of minor things:

1. The `osiApproved` field override in `SubmitLicenseSerializer` duplicates the choices that presumably exist in the model's `OSI_CHOICES`. A comment explaining why the override is needed (mixed int/string keys in the original) would help future readers understand this isn't accidental duplication -- I see the inline comment but it could be expanded slightly.

2. Removing `@permission_classes((AllowAny,))` from `submit_license` is correct since `AllowAny` is already the `DEFAULT_PERMISSION_CLASSES` in settings -- nice cleanup.

rtgdk (approved):
Approve. Excellent addition — the schema annotations are detailed, OAuth2/session auth is documented well, and the `_FORMAT_HELP` enum mapping is thoughtful. Concur with @rtgdk's approval.

A few non-blocking comments for follow-up:

1. **Doc/behavior mismatch on `/api2/`:** the PR description says "`/api2/` endpoints are also included", but `preprocess_exclude_internal_paths` in `src/api/spectacular.py` excludes both `/auth/` and `/api2/`. If `/api2/` is obsolete (per the linked commit comment in your earlier comment thread), excluding is the right call — just update the PR description to match.

2. **Public Swagger UI:** the Swagger UI is mounted at `/api/` with no auth check. Confirm this is intended (it usually is for a public API); just worth a sanity check that no internal-only endpoints could leak via the schema in the future.

3. **Cosmetic — PEP 8:** there are stray blank lines between the `@extend_schema(...)` / `@renderer_classes((JSONRenderer,))` decorator stack and the function `def` (e.g. in `validate`, `convert`, `compare`). Convention is no blank line between decorators and the function they decorate.

4. **Minor — `osiApproved` override comment:** agree with @rtgdk's note. A slightly expanded inline comment naming the model field with the mixed int/string keys would help future readers understand this isn't accidental duplication.

PR #646, opened 2026-04-27
ntia: Fix error popup

bact:
Error messages from NTIA Conformance Checker should be rendered, not being displayed as raw HTML tags.

Currently it shows HTML tags:

With fix from this PR:


goneall (approved):
LGTM

PR #647, opened 2026-04-30
Remove Unicode/encoding code from Python 2

bact:
Python 3 supports Unicode natively.

- Remove legacy encoding code that was necessary at the time but no longer now.
- Remove unnecessary `list()` for object that is already a list in Python 3.
- Specify 'rb' (read, binary) when sent text to XML processor.

### DIFF
--- a/src/api/admin.py
+++ b/src/api/admin.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # SPDX-FileCopyrightText: 2017 Rohit Lodha
 # Copyright (c) 2017 Rohit Lodha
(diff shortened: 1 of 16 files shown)


rtgdk (approved):
Clean Python 2 compatibility removal -- all changes are correct for Python 3.

A few highlights:
- Good catch adding the explicit `import re` in `api/utils.py` -- it was used at lines 45-48 without a direct import.
- The `codecs.open(...).read()` to `with open(...)` change in `beautify` also fixes a file handle leak.
- Opening XML files in `'rb'` mode for lxml is the right call.

LGTM.

goneall (approved):

PR #649, opened 2026-04-30
Update spdx-license-matcher to version 2.9

goneall:
This includes updates to JPype which is a dependency for PR #637


requirements.txt
bact:
```suggestion
jpype1==1.7.0
```

To pass the CI.
Since spdx-license-matcher requires jpype1==1.7.0

bact (approved):

PR #653, opened 2026-05-04
Make output in-line (no pop-ups)

### LINKED ISSUE #597: Improve the error message display for validate
The current pop-up is small and has a large font.

We could mirror the changes made to the conformance checker.

bact:
- Make all the result/error/warning messages displayed in the same app page
- No pop-ups
- Better font sizes
 
This will also improve web accessibility and allow screen reader to work better.

To fix #597

See screenshots below.


bact:
Screenshots from local test:

1) Validate, showing the validation status inline just below the "Validate" button; the file format will also be automatically selected based on file extension of the selected file



--


2) Convert, the download link will be below the "Convert" button



--


3) Compare, each selected file can now be removed individually



--

4) Conformance check, result is below the "Check conformance" button



--

5) Visual Editor (nothing change)



--

6) Check license, status is immediately below



--

7) License XML editor



--

8) License XML editor, editor view



--

9) Submit new license, retain the info pop-up



--

10) License requests list



--

11) Archived license requests list



--

12) About, 3 same-width columns, flow and stack by window size

goneall (commented):
I spent an hour or so attempting a review, but the changes are quite extensive.

It looks like there are some code formatting changes (e.g. removing spaces) mixed in with the more significant changes to in lining the modals.  There's also some changes to the style sheet which may or may not be related to the in lining.

If there is a way to separate out the code formatting from the significant changes, that would make the review much easier.

bact:
@goneall Thank you. I'll try to remove those formatting.

bact:
I will better break down this into smaller PRs.
But few of them may need to be merged in order

bact:
Use #655 instead.

PR #654, opened 2026-05-04
test: Use GitHub Actions to setup Firefox driver + use Chrome as fallback

bact:
- Use [browser-actions/setup-firefox](https://github.com/browser-actions/setup-firefox) to setup Firefox - much faster
- If Firefox is not available for test, try Chrome (already included in default GitHub runner)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,31 +56,13 @@ jobs:
           distribution: temurin
           java-version: "17"
       - name: Install Firefox
-        # Make sure to install the latest version of Firefox to avoid compatibility issues with geckodriver
+        # Use PPA to install non-snap Firefox, which is more reliable for Selenium on Ubuntu runners
         run: |
           sudo snap remove --purge firefox || true
           sudo add-apt-repository ppa:mozillateam/ppa -y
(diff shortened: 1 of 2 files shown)


rtgdk (approved):
Approve.

Switching to `browser-actions/setup-firefox@v1.7.1` (pinned by SHA — good practice) is a clean replacement for the snap-removal/PPA dance, and the workflow run is green. The refactor into `BaseSeleniumTestCase` with shared setUp/tearDown is solid; missing-driver is correctly handled via `skipTest`.

A couple of non-blocking comments:

1. The outer `@skipIf(not SELENIUM_AVAILABLE, ...)` decorators on the subclasses are now redundant since `BaseSeleniumTestCase.setUp` already calls `self.skipTest(...)` on the same condition. Either keep them (faster — skips before fixtures are touched) or drop them — the inconsistency between the two paths is worth a one-line comment in code or removal.

2. `print(f"Firefox initialization failed or geckodriver not found: {e}")` and the equivalent Chrome line at module import time will pollute stdout for every test session, including those that don't care about Selenium. Consider downgrading to `logging.warning(...)` (or guarding behind a verbose env var).

bact:
Thank you. I have update the code accordintly per review.

PR #655, opened 2026-05-06
Style: normalize fonts, cross-browser forms, inline result infrastructure

bact:
> [!NOTE]
> PR #653 is too big. It took me 5-6 hours to separate it out into smaller PRs as some of the stylesheets are interconnected and the results is not very good. So I asked Claude AI to give it a try. Still have to review in detailed + test manually locally, so I put all them as draft first).

This PR adjusts the base stylesheet to make consistent look & feel across all apps.
It sets the base font size for body text to 16 px as recommended by WCAG 2.2.

## Summary

Foundation PR — must merge before PRs #656, #657, #658, #659 (which rely on `showResult`/`hideResult`). (While #660 can be merged independently)

- Add inline result section CSS (`.result-section`, `.result-header`, `.result-body`) and shared JS helpers (`showResult` / `hideResult`) in `base.html` to replace Bootstrap modal popups on tool pages
- Normalize font sizes from `em` to `px` (16 px body baseline, WCAG 2.2 AA minimum)
- Reset form controls for cross-browser consistency (`-webkit-appearance: none` — fixes Safari native OS widget styles)
- Add custom SVG dropdown arrow for consistent `<select>` styling across browsers
- Add `.btn-submit-center`, `.form-narrow`, updated `.error-log`, DataTables search field overrides
- Fix `.btn-info` missing `color: #fff` (text was invisible on some browsers)
- Normalize font sizes in `dragdrop.css` and `editor/editor.css` to 16 px
- Update `starter-template.css`: left-align body content, adjust padding

## Files changed
- `src/app/templates/app/base.html`
- `src/app/static/css/dragdrop.css`
- `src/app/static/css/editor/editor.css`
- `src/app/static/css/starter-template.css`

## Test plan
- [x] All tool pages render without layout regressions
- [x] Form inputs and selects render consistently in Chrome, Firefox, Safari
- [x] `.btn-info` buttons show white text

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #659, opened 2026-05-06
Compare: individual file removal, auto filename, inline result

bact:
> [!NOTE]
> **Depends on #655** — review/merge that first.

## Summary

- Replace Bootstrap modal with inline result-section
- Show a numbered file list with per-file **Remove** button so users can drop individual files without clearing all selections (previously only "Reset all" was available)
- Auto-suggest result filename from first selected file's base name; user can override (tracked via `filenameUserEdited` flag)
- Duplicate file detection: adding the same file twice shows an error message
- Added `aria-label`, `aria-live`, `role="region"` throughout the form

## Files changed
- `src/app/templates/app/compare.html`

## Test plan
- [x] Add 3 files → numbered list with Remove buttons appears
- [x] Click Remove on file 2 → list renumbers, filename suggestion updates
- [x] Add a file twice → error message, file not added twice
- [x] Result filename auto-fills from first file, honours manual edits
- [x] Successful comparison → green result with download link
- [x] "Remove all" resets the list

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/TODO
+++ b/docs/TODO
@@ -1,6 +1,5 @@
 # TODO
 
-- Drag and drop file method.
 - TagToRdf takes max 3 arguments. Include output format field in forms and call method accordingly.
 - In api, create models function is saving the files again. So files sent to API is saved to 2 folders
     media/api/<username>/<time> --> This contains the result files too.
--- a/src/app/static/css/dragdrop.css
+++ b/src/app/static/css/dragdrop.css
@@ -22,13 +22,13 @@
 }
 #drop-area h4,
 #drop-area p {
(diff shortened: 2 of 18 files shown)


bact:
Warns if SPDX 3 JSON is selected.

Screenshot 2026-05-07 at 03 08 35

Compare successful, show download link.

Screenshot 2026-05-07 at 03 08 51

bact:
Changeset is now minimized. Ready for review.

goneall (approved):
LGTM

PR #662, opened 2026-05-07
Bump version to 1.4.0 in version.py

bact:
Bump version to 1.4.0 so it displayed correctly in About page


goneall (approved):
Good catch @bact

PR #671, opened 2026-05-09
Update README to use 'docker compose' commands

### LINKED ISSUE #666: Change README-PRODUCTION to use docker compose plugin rather than docker-compose
While updating production, the `docker-compose -f docker-compose.prod.yml ` command failed with an error missing key value for `containerConfig`.  Based on [this stackoverflow issue](https://stackoverflow.com/questions/78380867/docker-compose-run-issue-2024-error-containerconfig), it was caused by inconsistent docker-compose versions.

I was able to get the command working by using the `docker compose -f docker-compose.prod.yml ` (note the space instead of dash in `docker[-]compose`).

There may be other updates needed for the production environment - such as the installation method for docker-compose.

goneall:
Fixes #666

### DIFF
--- a/README-PRODUCTION.md
+++ b/README-PRODUCTION.md
@@ -59,18 +59,18 @@ Following are the steps for updating the images:
   - replace `<aws-account-id>` with the AWS account ID
   - replace `<aws-region>` with the AWS region
   - replace `<version>` with the specific version of the spdx-online-tools-build to be deployed
-- Build the image by running `docker-compose -f docker-compose.prod.yml build`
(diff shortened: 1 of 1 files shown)


rtgdk (approved):
Approve. Replaces deprecated `docker-compose` (v1, EOL Jun 2023) with `docker compose` (v2 plugin) and drops the manual v1 binary install instructions. Correct and overdue.

Optional nit (out of scope): `Recommend Ubuntu Server 22.04.2 LTS` could be revisited — 24.04 is the current LTS — but no need to block this PR on that.

bact (approved):

PR #677, opened 2026-06-23
Include issue tracker URL in the error message

### LINKED ISSUE #157: Improve error message when error received from the Java code
If there is an unexpected exception from the Java Bridge, add some text before the message to describe the error and action.  For example "An unexpected error occurred.  Please submit an issue at github.com/spdx/spdx-online-tools/issues with the following error: "

See related SPDX tools issue https://github.com/spdx/tools/issues/229

bact:
When there is an error, tell user how to report it.

Also indicate if the error comes from Java code.

To fix #157

### DIFF
--- a/src/app/core.py
+++ b/src/app/core.py
@@ -21,6 +21,17 @@ from spdx_license_matcher.utils import get_spdx_license_text
 
 import app.utils as utils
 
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - Thanks @bact

PR #674, opened 2026-05-11
Use semantic HTML for better accessibility

bact:
Using semantic HTML structure to give hint to screen reader and keyboard navigation.

- Title of the app should be in h1 heading.
- Sections go to h2, h3, etc.

PR #683, opened 2026-06-24
Upgrade to Django 6.0

bact:
Upgrade Django to version 6.0 as version 5.2 is end of mainstream support since Dec 2025.

6.0 mainstream support will end August 2026 which by the time we will have 6.1.


rtgdk (approved):
Reviewed the diff. Straightforward upgrade to Django 6.0 with the Python floor moved to 3.12 (required by Django 6.0), consistent updates across the Dockerfiles, README, and settings doc links. The `social-auth-app-django` / `social-auth-core` pins are well documented — the inline comments explaining the `GooglePlusAuth` import constraint from `drf-social-oauth2` are exactly the kind of note that saves the next person a debugging session. CI is green on 3.12.

One coordination note, not a blocker: the CI matrix drops 3.10 and now tests only 3.12, while the other open PRs (#674, #684, #685, #686) still target 3.10. If this lands first, those branches will need a rebase to pick up the new matrix; if they land first, this one is unaffected. Just worth sequencing deliberately.

Approving.

PR #685, opened 2026-06-26
License info page: Add link to GitHub issue

### LINKED ISSUE #453: Add link of the GitHub Issue to the License Information page
Even though there is a link to access the request of the online tool from the Github issue, there is no link to access the GitHub issue from the online tool.

bact:
- To resolve #453
- This may make it more convenient for license reviewer.
- With this PR, GitHub issue number is added, with link (screenshot with mock data):

### DIFF
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -539,7 +539,10 @@ def submit_license(request):
             if 'urlType' in request.POST:
                 # This is present only when executing submit license via tests
                 urlType = request.POST["urlType"]
-            statusCode, githubIssueId = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
+            statusCode, githubIssueId, githubIssueUrl = createIssue(licenseAuthorName, licenseName, licenseIdentifier, licenseComments, licenseSourceUrls, licenseHeader, licenseOsi, licenseRequestUrl, token, urlType)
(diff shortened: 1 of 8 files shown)


bact:
Since we are going to insert the GitHub issue+link to the table, this PR also fixes the misaligned buttons at the bottom of the license information page, which currently looks jumpy like this:



(screenshot taken from https://tools.spdx.org/app/license_requests/507/ )

rtgdk (approved):
Reviewed the diff. The change is clean: new `github_issue_url` / `github_issue_number` fields with a matching migration, `createIssue` extended to return the issue's `html_url`, and the license information page rendering a link when present. I verified all four `createIssue` call sites (three in `app/views.py`, one in `api/views.py`) unpack the new 3-tuple, and both return paths inside `createIssue` yield three values, so no caller is left on the old 2-tuple contract.

Worth calling out that the `api/views.py` `submit_license` call now passes `[]` for `licenseExamples`. Previously that argument was omitted, which silently shifted every positional argument after it — so this also fixes a pre-existing latent bug in that path. Good catch.

One coordination note: this PR and #674 both restructure `src/app/templates/app/license_information.html` substantially (this one wraps fields in `.licenseInfoTable` / `.licenseInfoField`; #674 converts them to a `<dl>`). They will conflict on merge, so whichever lands second will need a rebase, and the GitHub-issue link block added here should be carried into the final structure.

Approving.

bact:
The test is now put properly in the new tests/ dir layout (introduced by #684) (inside test_legacy.py, wait to be refactor further in next PRs)

PR #686, opened 2026-06-26
Fix large file upload issue

### LINKED ISSUE #499: Large file uploads to API failing with FileNotFoundError
Querying the API (e.g. /api/validate/) with a file larger than (I believe) 2.5 MB results in an error similar to `[Errno 2] No such file or directory: '/tmp/tmpz4g1gbzc.upload.json'` in the `[Function]FileUpload.objects.create()` model calls.

I believe this has to do with Django storing sufficiently large files in a temp location, unlike small files which are stored in memory. It appears that these files are opened twice in most of the API routes, but have already been deleted by the second open. I have fixed this on my local instance by, in the helpers like `license_validate_helper` in /app/core.py, returning the permanently saved `filename` as part of the response. Then in the API routes like `validate` in /api/views.py, the `query` object is created using `file=response['filename']` rather than trying to get the file again from `request.FILES['file']`.

If you agree with the issue and solution and would like me to send a PR then I will, I was just unsure about your stylistic best practices and if I was the only one experiencing this.

bact:
When uploading file larger than 2.5 MB, the online tools will failed with "file not found" error.

Root cause
- Django stores uploads >2.5 MB as a TemporaryUploadedFile on disk.
- The core helpers (license_validate_helper, license_convert_helper, license_compare_helper) move that temp file into the media directory on first save.
- The API views try to do another save of the same upload object, but the file is no longer there -- failed.

Records the upload path in `result['uploaded_file']` and let views use that path directly (instead of re-saving from no longer existing location)

To fix #499 as reported and analysed root cause by @NiViggiano

### DIFF
--- a/src/api/test_upload.py
+++ b/src/api/test_upload.py
@@ -1,0 +1,144 @@
+# SPDX-FileCopyrightText: 2026 SPDX contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for large-file uploads (issue #499).
+
+Django stores uploads above FILE_UPLOAD_MAX_MEMORY_SIZE as a TemporaryUploadedFile
+on disk. The core helper moves that temp file into media on first save, so a second
+save in the view raised FileNotFoundError. Setting the threshold to 0 forces every
+upload through that path without needing real multi-MB fixtures.
+"""
+
+import os
+
+from django.conf import settings
+from django.contrib.auth.models import User
+from django.core.exceptions import ObjectDoesNotExist
+from django.test import override_settings
+from django.urls import reverse
+from rest_framework.test import APITestCase
+
+from api.models import CompareFileUpload, ConvertFileUpload, ValidateFileUpload
+
+
+def getExamplePath(filename):
+    return os.path.join(settings.EXAMPLES_DIR, filename)
+
+
+@override_settings(FILE_UPLOAD_MAX_MEMORY_SIZE=0)
+class ValidateLargeFileUploadTests(APITestCase):
+    """Validate API must handle uploads routed through TemporaryUploadedFile."""
+
+    def setUp(self):
+        self.username = "validateuploadtestuser"
+        self.password = "validateuploadtestpass"
+        self.tearDown()
+        self.credentials = {"username": self.username, "password": self.password}
+        u = User.objects.create_user(**self.credentials)
+        u.is_staff = True
+        u.save()
+
+    def tearDown(self):
+        try:
+            u = User.objects.get_by_natural_key(self.username)
+            u.delete()
+        except ObjectDoesNotExist:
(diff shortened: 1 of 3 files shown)


bact:
@NiViggiano Thanks for the fix and if you have time, please help review

bact:
Note that the tests are intentionally put in a separated `test_upload.py` and not inside `tests.py` to avoid conflicts with #684.

`python src/manage.py test api app` command will still discover the tests.

rtgdk (approved):
Reviewed the diff. The root-cause analysis is correct: Django stores uploads above `FILE_UPLOAD_MAX_MEMORY_SIZE` as a `TemporaryUploadedFile`, and the second save in the API views was failing because the helper had already moved that temp file into media. Recording the saved path in the helper result and reusing it in the views is the right fix, and forcing the threshold to 0 in the regression tests is a nice way to exercise the on-disk path deterministically. Nicely scoped.

A couple of minor, non-blocking notes:

1. Compare file ordering. `license_compare_helper` builds `saved_files` by iterating `request.FILES.getlist("files")`, and `compare()` then assigns `query.file1.name = saved_files[0]` / `query.file2.name = saved_files[1]`. This relies on `getlist("files")` returning file1 before file2. That happens to hold today, but the coupling is implicit — a short comment (or mapping the saved paths back by field name) would make it robust if the form field order ever changes.

2. Test file modes. `test_validate_api_large_file` and `test_compare_api_large_file` open the example files in text mode, while `test_convert_api_large_file` uses `"rb"`. Harmless here, but worth making consistent.

3. Optional: a one-line comment on `query.file.name = saved_file` noting that the file is already persisted on disk by the helper (so this deliberately avoids a re-save) would help future readers.

Approving — the notes above are optional polish.

bact:
> A couple of minor, non-blocking notes:
> 
>     1. Compare file ordering. `license_compare_helper` builds `saved_files` by iterating `request.FILES.getlist("files")`, and `compare()` then assigns `query.file1.name = saved_files[0]` / `query.file2.name = saved_files[1]`. This relies on `getlist("files")` returning file1 before file2. That happens to hold today, but the coupling is implicit — a short comment (or mapping the saved paths back by field name) would make it robust if the form field order ever changes.
> 
>     2. Test file modes. `test_validate_api_large_file` and `test_compare_api_large_file` open the example files in text mode, while `test_convert_api_large_file` uses `"rb"`. Harmless here, but worth making consistent.
> 
>     3. Optional: a one-line comment on `query.file.name = saved_file` noting that the file is already persisted on disk by the helper (so this deliberately avoids a re-save) would help future readers.
> 
> 
> Approving — the notes above are optional polish.

Thank you. I will address those in next PR.
Will merge this and then #683 first, to avoid conflicts with other future PRs.

PR #693, opened 2026-07-11
Organize tests - Lot 0 - Move all tests to a dedicated tests/ directory

bact:
- Consolidate all API and App tests, including helpers, to be under one `src/tests` test root
- Add `src/tests/` to `.dockerignore` to exclude tests and mock files from production Docker images

Apart from add new `__init__.py` for the new tests dir. This PR is a pure move.


rtgdk (approved):
Clean, pure file moves with zero logic changes. The `src/tests/` addition to `.dockerignore` and the CI command update are both correct. Ready to merge.

PR #698, opened 2026-07-13
Organize tests - Lot 7 - Move validate, convert, compare API tests

bact:
Migrate tests for the Validate, Convert, and Compare APIs
from one huge `test_api.py` file
to dedicated `test_validate_api.py`, `test_convert_api.py`, and `test_compare_api.py` files.

Also consolidates the large-file upload tests for the Validate, Convert, and Compare APIs from `test_upload.py` into above files.

Split tests the Validate, Convert, and Compare apps
from one `test_tool.py` file
to dedicated `test_validate.py`, `test_convert.py`, and `test_compare.py` files.

Migrated classes:

- `ValidateFileUploadTests`
- `ValidateLargeFileUploadTests`
- `ConvertFileUploadTests`
- `ConvertLargeFileUploadTests`
- `CompareFileUploadTests`
-  `CompareLargeFileUploadTests`


rtgdk (commented):
**Review summary**

Two confirmed correctness bugs (easy fixes) and two plausible issues were found. All pre-exist in the original `test_api.py`/`test_upload.py` — this reorganization didn't introduce them, but it's a good moment to clean them up.

**Verdict: Conditional approval** — the reorganization is a clear improvement. Fix the `assertTrue` calls before merging (one-liner per site). The EOF file-handle bug is also worth fixing here since the code is being touched; the remaining two are lower priority.

src/tests/api/test_validate_api.py
```diff
@@ -0,0 +1,144 @@
...
+            u = User.objects.get_by_natural_key(self.username)
+            u.delete()
+        except ObjectDoesNotExist:
+            pass
+        ValidateFileUpload.objects.all().delete()
+
+    def test_validate_api(self):
+        """Access get without login"""
+        resp1 = self.client.get(reverse("validate-api"))
+        self.assertTrue(resp1.status_code, 403)
```
rtgdk: **Bug (CONFIRMED): `assertTrue(expr, msg)` — status code is never actually checked**

`self.assertTrue(resp1.status_code, 403)` treats `403` as the *failure message*, not an expected value. It only checks `bool(resp1.status_code)`, which is `True` for any non-zero HTTP code. A broken endpoint returning 500 silently passes this assertion.

Same issue on line 55 here, and in `test_compare_api.py` lines 51/55 and `test_convert_api.py` lines 54/58 — 6 call sites total across all three new files.

```python
# Fix — replace assertTrue with assertEqual everywhere
self.assertEqual(resp1.status_code, 403)
self.assertEqual(resp2.status_code, 200)
```

src/tests/api/test_compare_api.py
```diff
@@ -0,0 +1,148 @@
...
+                "file1": self.rdf_file,
+                "file2": self.tv_file,
+                "rfilename": "compare-apitest.xls",
+            },
+            format="multipart",
+        )
+        self.assertEqual(resp4.status_code, 400)
+        self.client.logout()
+        self.tearDown()
+
```
rtgdk, line 87: **Bug (CONFIRMED): `self.rdf_file` is at EOF here — resp4 uploads 0 bytes**

`setUp` opens `self.rdf_file` once. The `resp3` POST above called `file.read()` on it to build the multipart body, advancing the cursor to EOF. This second POST therefore sends a zero-byte `file1`.

The test asserts `status 400` and gets it — but because the server rejected an *empty file*, not because it detected the RDF-vs-TAG format mismatch the comment claims. The format-validation path is never actually exercised.

```python
# Fix — seek back before the second POST
self.rdf_file.seek(0)
resp4 = self.client.post(reverse("compare-api"), {"file1": self.rdf_file, ...})
```

(Predates this PR; was copied from the original `test_api.py`.)

src/tests/api/test_validate_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def setUp(self):
+        self.username = "validateapitestuser"
+        self.password = "validateapitestpass"
+        self.tearDown()
+        self.credentials = {"username": self.username, "password": self.password}
+        u = User.objects.create_user(**self.credentials)
+        u.is_staff = True
+        u.save()
+        self.tv_file = open(getExamplePath("SPDXTagExample-v2.0.spdx"))
```
rtgdk, line 35: **Issue (PLAUSIBLE): File handles opened in `setUp` are never closed in `tearDown`**

`self.tv_file`, `self.rdf_file`, etc. are instance attributes that persist until the test instance is GC'd. `tearDown` deletes DB rows but never calls `.close()`. On CPython this resolves via refcounting (no practical leak), but it triggers `ResourceWarning` under `-Wd` and is a real leak under PyPy/Jython. Same pattern in `test_compare_api.py` and `test_convert_api.py`.

```python
# Fix — register cleanup alongside the open call
self.tv_file = open(getExamplePath("SPDXTagExample-v2.0.spdx"))
self.addCleanup(self.tv_file.close)
```

src/tests/api/test_validate_api.py
```diff
@@ -0,0 +1,144 @@
...
+            {"file": self.invalid_rdf_file, "format": "RDFXML"},
+            format="multipart",
+        )
+        self.assertEqual(
+            resp6.data["owner"], User.objects.get_by_natural_key(self.username).id
+        )
+        self.assertEqual(resp6.status_code, 400)
+        self.assertNotEqual(resp6.data["result"], "This SPDX document is valid.")
+        self.client.logout()
+        self.tearDown()
```
rtgdk: **Issue (PLAUSIBLE): Explicit `tearDown()` call is redundant — Django's runner calls it automatically**

The test runner calls `tearDown` after every test method, so this explicit call makes `tearDown` fire twice per test. The current implementation is idempotent (`ObjectDoesNotExist` is caught, `.delete()` on an empty queryset is a no-op), so there's no failure today. But any future non-idempotent addition to `tearDown` (an assertion, counter, resource release) will silently double-fire.

Same pattern appears at the end of `test_validate_without_argument`, `test_compare_api`, `test_compare_without_one_argument`, and the large-file tests. Remove all explicit `self.tearDown()` calls at the end of test methods and let the runner handle it.

src/tests/api/test_compare_api.py
```diff
@@ -0,0 +1,149 @@
...
+                "rfilename": "compare-apitest.xls",
+            },
+            format="multipart",
+        )
+        self.assertEqual(resp3.status_code, 200)
+        self.assertEqual(
+            resp3.data["owner"], User.objects.get_by_natural_key(self.username).id
+        )
+        self.assertTrue(resp3.data["result"].startswith(settings.MEDIA_URL))
+        """Compare with one  invalid RDF files"""
```
rtgdk: Nit: this comment is now stale. After the `seek(0)` fix, both files are valid SPDX documents and the API returns 200 — not an error case.

```python
# Before (stale):
"""Compare with one  invalid RDF files"""

# After:
"""Compare RDF file with TAG/TV file (mixed formats, both valid SPDX)"""
```

rtgdk (approved):
All previous findings are resolved. One nit remains inline.

bact:
Stale comment updated as suggested.

PR #701, opened 2026-07-14
Organize config + management command dirs

bact:
- **Config relocation**: Renamed the configuration package `src/src/` to `src/config/`, make it less confusing and has more meaningful name; updating related code, tests, workflow, and documentation.
- **Cleanup util relocation**: Converted `src/app/utils.py` into a package and moved the media cleanup script from `app/scripts/cleanup.py` to `app/utils/cleanup_media.py` (matching the command name); rename `test_cron.py` to `test_cleanup_media.py`.

Follow-up of test suite reorganization.


src/config/wsgi.py
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2017-present SPDX Contributors
+# SPDX-FileType: SOURCE
+# SPDX-License-Identifier: Apache-2.0
+
 """
 WSGI config for project src.
```
rtgdk: **Nit:** docstring still says `project src` after the rename — should be `project config` (or drop the project name entirely, as it's now encoded in the file path).

```python
"""WSGI config for project config."""
```

rtgdk (approved):
## Review summary

Clean, thorough rename. All 26 files are consistently updated — no remaining `from src.*` imports, no leftover `app.scripts.cleanup` references, `supervisor_api.conf` and `manage.py` are both correct, and the `utils` package is properly initialized with `__init__.py`. `README-PRODUCTION.md` and the Dockerfile are unaffected (no Dockerfile exists). One cosmetic nit in `wsgi.py`; nothing blocking.

**Verdict: Approved. Ready to merge.**

src/config/wsgi.py
bact: Thank you. Done.

Issue #700, opened 2026-07-14
Add / complete SPDX headers of files in repo


bact:
As one of many steps to achieve https://github.com/spdx/spdx-online-tools/issues/209

Should have at least these three fields:

```python
# SPDX-FileCopyrightText:
# SPDX-FileType:
# SPDX-License-Identifier:
```

PR #704, opened 2026-07-15
Update NOTICE and  SPDX-License-Identifier headers

bact:
For code / dependency from
- CodeMirror (MIT)
- jsdifflib (BSD-3-Clause and BSD-2-Clause)
- SPDX XML license formatter (MIT)
- SPDX Tools (tools-java) (Apache-2.0) + components bundled in tool.jar (with BSD-3-Clause components)

As part of #700 effort and follows Section 4(d) of the Apache License 2.0.

Use Apache Arrow NOTICE as example
https://github.com/apache/arrow/blob/main/NOTICE.txt

Add `LICENSES/` directory to include all license texts (Apache-2.0, BSD-2-Clause, BSD-3-Clause, MIT).

Modify existing Docker files to use `_license-list-data/` instead of `licenses/`, to avoid conflicts with `LICENSES/` (seen the same in case-insensitive file system like on macOS).


rtgdk (commented):
The NOTICE restructure and SPDX normalisation are both real improvements — the structural cleanup is welcome and fixing the missing `SPDX-License-Identifier` in `difflib.js` is a genuine gap closed.

Main concern before merging: the full MIT and BSD-3-Clause license texts were removed from NOTICE. Both licenses require their copyright notice **and** permission notice to be included in all copies. SPDX identifiers in source files satisfy the copyright-notice part, but not the full permission-notice boilerplate. If there is no `LICENSES/` directory in the repo, consider:
- Adding `LICENSES/MIT.txt` and `LICENSES/BSD-3-Clause.txt` (REUSE-compliant layout), or
- Restoring the full texts under each component section in NOTICE.

NOTICE
```diff
@@ -1,76 +1,73 @@
...
- * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
- * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
- * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
- * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
- * DAMAGE.
- */
-
- Additional open source software dependencies are listed in the requirements.txt file
+SPDX Online Tools
+Copyright 2017-present SPDX Contributors
```
rtgdk: Nit: consider an explicit year range like `2017-2026` instead of `2017-present` — some SBOM and compliance tooling cannot parse `present` as a date.

NOTICE
```diff
@@ -1,76 +1,73 @@
...
+
+This product includes software developed by the SPDX Contributors
+(https://spdx.dev/).
+
+=========================================================================
+
+This product includes software from the CodeMirror project (MIT).
```
rtgdk, line 9: The full MIT license text was removed here. MIT requires:
> The above copyright notice and **this permission notice** shall be included in all copies.

SPDX identifiers in source files cover the copyright notice, but not the full permission-notice text. Consider adding `LICENSES/MIT.txt` (REUSE layout) or restoring the full text here.

NOTICE
```diff
@@ -1,76 +1,73 @@
...
+Copyright (C) 2017 by Marijn Haverbeke <marijnh@gmail.com> and others.
+https://codemirror.net/
+
+The files with the included software are located in:
+- src/app/static/css/editor/codemirror.css
+- src/app/static/js/editor/codemirror.js
+
+-------------------------------------------------------------------------
+
+This product includes software from the jsdifflib project (BSD-3-Clause).
```
rtgdk, line 19: Same concern as for CodeMirror above — the full BSD-3-Clause text was also removed. BSD-3-Clause has equivalent requirements. Consider `LICENSES/BSD-3-Clause.txt`.

src/app/formatxml.py
```diff
@@ -1,11 +1,13 @@
...
-#
+
+# SPDX-FileContributor: Alexios Zavras
 # SPDX-FileCopyrightText: 2017 Alexios Zavras
-# Copyright (c) 2017 Alexios Zavras
+# SPDX-FileType: SOURCE
 # SPDX-License-Identifier: MIT
 
+# quick-n-dirty formatter for SPDX licenses in XML format
+# Copyright (c) 2017 Alexios Zavras
```
rtgdk: Nit: this is now redundant with `SPDX-FileCopyrightText: 2017 Alexios Zavras` on line 4. Keeping both is confusing in an SPDX tooling project; the legacy line can be removed.

src/app/static/css/editor/codemirror.css
```diff
@@ -1,3 +1,11 @@
+/*
+ * SPDX-FileContributor: Marijn Haverbeke <marijnh@gmail.com>
+ * SPDX-FileCopyrightText: CodeMirror, copyright (c) by Marijn Haverbeke and others
```
rtgdk: The SPDX spec recommends `SPDX-FileCopyrightText: <year> <name>`. This value embeds the word "copyright" and omits a year. `codemirror.js` carries the same pre-existing header — consider normalising both to:
```
SPDX-FileCopyrightText: 2017 Marijn Haverbeke and others
```

src/app/static/css/editor/diffview.css
```diff
@@ -1,3 +1,10 @@
+/*
+ * SPDX-FileContributor: Chas Emerick <cemerick@snowtide.com>
+ * SPDX-FileCopyrightText: Copyright 2007 - 2011 Chas Emerick <cemerick@snowtide.com>
+ * SPDX-FileType: SOURCE
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
 /*
 This is part of jsdifflib v1.0. <http://github.com/cemerick/jsdifflib>
```
rtgdk: This legacy attribution comment uses `http://` while NOTICE now consistently uses `https://`. It is also redundant with the SPDX header block above — consider removing it or at least updating the URL scheme.

NOTICE
bact: Thank you. Done.

src/app/formatxml.py
bact: Done.

src/app/static/css/editor/codemirror.css
bact: Done.

src/app/static/css/editor/diffview.css
bact: Updated to https

src/app/static/css/editor/codemirror.css
```diff
@@ -1,3 +1,12 @@
+/*
+ * SPDX-FileContributor: Marijn Haverbeke <marijnh@gmail.com>
+ * SPDX-FileCopyrightText: 2017 Marijn Haverbeke and others
+ * SPDX-FileType: SOURCE
+ * SPDX-License-Identifier: MIT
+ */
+
+/* This CSS is from CodeMirror (http://codemirror.net) */
```
rtgdk: Nit: `http://` here while the rest of this PR moves to `https://`. Consider `https://codemirror.net`.

src/app/static/js/editor/difflib.js
```diff
@@ -1,5 +1,7 @@
-// SPDX-FileCopyrightText: Copyright (c) 2007, Snowtide Informatics Systems, Inc. All rights reserved.
 // SPDX-FileContributor: Chas Emerick <cemerick@snowtide.com>
+// SPDX-FileCopyrightText: 2007, Snowtide Informatics Systems, Inc.
```
rtgdk: Nit: SPDX `SPDX-FileCopyrightText` format is `<year> <name>` (space-separated). The comma after `2007` makes this non-standard — `2007 Snowtide Informatics Systems, Inc.`

rtgdk (approved):
All concerns from the first round addressed — the REUSE-compliant `LICENSES/` restructure and the `_license-list-data/` rename are nice extras. Two small nits inline, neither blocking.

NOTICE
bact: jsdifflib source fule contain full license text.

Additional license text is now provided at LICENSES/ directory, per REUSE spec

bact:
@rtgdk I have found that while [`src/app/static/js/editor/difflib.js`](https://github.com/spdx/spdx-online-tools/blob/main/src/app/static/js/editor/difflib.js) uses BSD-3-Clause, [`src/app/static/css/editor/diffview.css`](https://github.com/spdx/spdx-online-tools/blob/main/src/app/static/css/editor/diffview.css) and [`src/app/static/js/editor/diffview.js`](https://github.com/spdx/spdx-online-tools/blob/main/src/app/static/js/editor/diffview.js) license declaration are actually BSD-2-Clause (has only 2 clauses in the license text in each of the file).

With that finding, I fixed SPDX-License-Identifier in the headers and add BSD-2-Clause.txt to `LICENSES/` directory.

NOTICE
```diff
@@ -1,76 +1,80 @@
...
+Copyright (c) 2007, Snowtide Informatics Systems, Inc.
+https://github.com/cemerick/jsdifflib
+
+The file with the included software is located in:
+- src/app/static/js/editor/difflib.js
+
+-------------------------------------------------------------------------
+
+This product includes software from the jsdifflib project (BSD-2-Clause).
```
rtgdk, line 28: Can you confirm the BSD-2-Clause assignment was verified against the actual `LICENSE` file in the [cemerick/jsdifflib](https://github.com/cemerick/jsdifflib) repo, rather than inferred from the copyright year range? The original Snowtide code was BSD-3-Clause, so the drop of the non-endorsement clause for Emerick's fork is the key thing to nail down.

rtgdk (approved):
Latest round looks good — the BSD-2/3-Clause split is a careful finding and the https cleanups in `codemirror.js` are a nice bonus. One question inline on the license split.

NOTICE
bact: @rtgdk  I can confirm that the BSD-2-Clause assignment was verified against the actual license headers in `diffview.css` and `diffview.js`.

The repo doesn't have the LICENSE or any license information.

But the source files contain clear license text:

- https://github.com/cemerick/jsdifflib/blob/master/difflib.js#L1-L19 (BSD-3-Clause)
- https://github.com/cemerick/jsdifflib/blob/master/diffview.css#L1-L16 (BSD-2-Clause)
- https://github.com/cemerick/jsdifflib/blob/master/diffview.js#L1-L16 (BSD-2-Clause)

rtgdk: Thanks for checking — sourcing it from the in-file license text when there's no standalone LICENSE file is exactly the right call. Confirmed, this looks good.

rtgdk (approved):
Confirmed — BSD-2/3-Clause split is well-sourced. This is ready to merge.

Issue #472, opened 2023-06-02
Add linters to the project


BassCoder2808:
Linters are tools that help identify and report errors in code, helping to ensure that the codebase is 
of a high quality and follows best practices. The SPDX Online Tools project could benefit from the 
addition of linters for Python, JavaScript, HTML, and CSS, which would help to identify and fix issues 
in the codebase.

Python linters such as Pylint, Flake8, and Pyflakes can help to identify errors, potential bugs, and 
code smells in Python code. JavaScript linters like ESLint can detect issues such as unused variables, 
missing semicolons, and incorrect variable scoping. HTML linters like HTMLHint can check for issues 
such as invalid markup and accessibility errors. Finally, CSS linters like Stylelint can help identify 
errors such as invalid selectors, unused styles, and syntax errors.

The linters could be integrated into the development process, either through the use of pre-commit 
hooks or as part of a continuous integration (CI) pipeline. When a developer commits code, the 
linters would run and report any issues. This would help to catch errors early in the development 
process, before they become larger problems.

bact:
It is a good idea to have linters for for Python, JavaScript, HTML, and CSS.

The issue is how to deal with the existing code?

Or will it be acceptable to run the auto-formatters for all languages at once and submit a huge PR (or few of them)? The PR that will do only one thing, formatting, and we trust the committer and our test infrastructure (because it will be labor-intensive to manually review it).

If that PR/review plan is acceptable, I think we can add the linters on GitHub workflow.

goneall:
This many changes would cause a lot of re-work on the existing PR's - I suggest we hold off on this until the other PR's are merged

PR #705, opened 2026-07-20
Add minimal linter and format check for Python tests

bact:
Limited the Python lint and format checks only for

- `src/*.py` (excluding subdirs) = 2 files
- `src/tests/*.py` (excluding subdirs) = 1 file
- `src/tests/api/` (including subdirs) = 3 files

Other files are unaffected.

Can gradually expand later when ready.
For example, expand to the small `src/config` dir in another PR.
The `pylint_per_file_ignores` plugin allows to set "lint ignores" per file.

First step to address #472

### DIFF
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -1,4 +1,4 @@
-name: Python formatter
+name: Python format
 
 on:
   push:
@@ -24,7 +24,7 @@ permissions:
   contents: read
 
 jobs:
-  lint:
+  black:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -37,4 +37,4 @@ jobs:
       - name: Black Python code formatter
         uses: psf/black@87928e6d6761a4a6d22250e1fee5601b3998086e  # v26.5.1
         with:
-          src: "src/*.py src/tests/*.py src/tests/api"
+          src: "src"
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -1,0 +1,40 @@
+name: Python formatter
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/*.py'
+      - 'src/tests/*.py'
+      - 'src/tests/api/**/*.py'
+      - 'pyproject.toml'
+      - '.github/workflows/python-format.yml'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'src/*.py'
+      - 'src/tests/*.py'
+      - 'src/tests/api/**/*.py'
+      - 'pyproject.toml'
+      - '.github/workflows/python-format.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6.3.0
+        with:
+          python-version: "3.12"
+      - name: Black Python code formatter
+        uses: psf/black@87928e6d6761a4a6d22250e1fee5601b3998086e  # v26.5.1
+        with:
+          src: "src/*.py src/tests/*.py src/tests/api"
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -29,9 +29,9 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
       - name: Set up Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6.3.0
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
         with:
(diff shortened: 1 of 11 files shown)


.github/workflows/python-format.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
+        with:
+          python-version: "3.12"
+      - name: Black Python code formatter
+        uses: psf/black@87928e6d6761a4a6d22250e1fee5601b3998086e  # v26.5.1
+        with:
+          src: "src"
```
rtgdk: **Scope mismatch:** The workflow triggers only on `src/*.py`, `src/tests/*.py`, and `src/tests/api/**/*.py` (lines 8–10), but Black is run against the entire `src: "src"` tree. Your PR description says *"Other files are unaffected"* — but if a file in e.g. `src/api/` or `src/tests/app/` fails Black's check, this job will fail whenever any in-scope change triggers it, even though those files were never part of the stated scope.

Consider scoping Black's input to match the trigger paths:
```yaml
with:
  src: "src/manage.py src/populate.py src/tests"
```
Or ensure the `pyproject.toml` `exclude` block fully covers every out-of-scope path.

.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,47 @@
...
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: ["3.12"]
```
rtgdk: **Unnecessary single-element matrix:** A matrix with a single entry (`["3.12"]`) adds YAML indirection and matrix-evaluation overhead with no benefit. Simplify to a direct value:
```yaml
    steps:
    - name: Set up Python
      uses: actions/setup-python@...
      with:
        python-version: "3.12"
```
The `${{ matrix.python-version }}` references on lines 36 and 39 would become the literal `"3.12"` directly.

pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+)
+'''
+
+[tool.pylint.main]
+load-plugins = ["pylint_per_file_ignores"]
+
+[tool.pylint."messages control"]
+per-file-ignores = [
+    "src/tests/*:invalid-name,attribute-defined-outside-init,duplicate-code,no-member,missing-module-docstring,missing-class-docstring,missing-function-docstring,unspecified-encoding,consider-using-with,too-many-instance-attributes,super-with-arguments,too-few-public-methods,not-callable,broad-exception-caught,too-many-return-statements,line-too-long,consider-using-f-string,missing-timeout",
+    "src/tests/**/*:invalid-name,attribute-defined-outside-init,duplicate-code,no-member,missing-module-docstring,missing-class-docstring,missing-function-docstring,unspecified-encoding,consider-using-with,too-many-instance-attributes,super-with-arguments,too-few-public-methods,not-callable,broad-exception-caught,too-many-return-statements,line-too-long,consider-using-f-string,missing-timeout",
```
rtgdk: **Redundant duplicate patterns:** Both entries carry an identical 18-rule ignore list. In most glob implementations (including `pylint-per-file-ignores`), `src/tests/**/*` is a superset of `src/tests/*`, making the first entry redundant.

More critically, maintaining two copies is fragile — a future edit to one will silently diverge from the other. Unless you have confirmed that both patterns are required for this plugin, keep only `src/tests/**/*`.

requirements-dev.txt
```diff
@@ -1 +1,4 @@
+black>=26.5.1
+pylint>=4.0.6
```
rtgdk: **Loose version pins:** Both `black` and `pylint` use lower-bound-only constraints. Both tools can introduce new formatting rules or lint checks in minor/major releases that would break CI without any code change in this repo.

Consider adding upper bounds:
```
black>=26.5.1,<27
pylint>=4.0.6,<5
```
Note that `selenium` already uses `<5.0.0` — worth applying the same discipline to the linting tools.

src/populate.py
```diff
@@ -5,38 +5,62 @@
...
+        if created:
             new_count += 1
-        if type=="licenses":
-            LicenseNames.objects.get_or_create(name=i["licenseId"])
+        if item_type == "licenses":
+            # pylint: disable=no-member
+            LicenseNames.objects.get_or_create(name=item["licenseId"])
         else:
-            LicenseNames.objects.get_or_create(name=i["licenseExceptionId"])
+            # pylint: disable=no-member
```
rtgdk: **Nit — three repeated `no-member` disables for the same call pattern:** `LicenseNames.objects.get_or_create` is the only call driving this suppression, yet the disable appears three separate times. A single disable scoped to the function or the loop would cover all three without the repetition:

```python
for item in data[item_type]:
    total_count += 1
    # pylint: disable=no-member  # Django ORM; .objects not visible to static analysis
    created = LicenseNames.objects.get_or_create(name=item["name"])[1]
    ...
    LicenseNames.objects.get_or_create(name=item["licenseId"])
    ...
    LicenseNames.objects.get_or_create(name=item["licenseExceptionId"])
```

Alternatively, add `populate.py` to the `per-file-ignores` list in `pyproject.toml` under the `no-member` rule.

.github/workflows/python-format.yml
bact: fixed

rtgdk (approved):
All five previous comments have been addressed — thank you for the quick turnaround.

One minor scope asymmetry to be aware of as you expand coverage in future PRs: **Pylint** targets `src/tests/*.py` + `src/tests/api` explicitly, while **Black** walks the full `src/tests/` directory (with `src/tests/app/` and `src/tests/fixtures/` excluded via `pyproject.toml`). This is consistent with the gradual rollout intent stated in the PR description, but worth noting so the next expansion PR knows where the gap is.

Approving — clean, well-scoped first step toward #472.

PR #710, opened 2026-08-13
Use license-matcher from PyPI

bact:
Use the SPDX License Matcher (`license-matcher`) from PyPI, similar to other libraries.


goneall (approved):
LGTM - Thanks @bact !

PR #718, opened 2026-08-18
Speed up test CI - use cache

bact:
- Use pip cache to avoid re-downloads
- Cancel jobs from previous commits in the same PR, so the most recent job can get pick up quicker
- Enable parallel for API tests / No parallel for app tests because parallel with browser sessions can be flaky

### DIFF
--- a/.github/workflows/test-api.yml
+++ b/.github/workflows/test-api.yml
@@ -36,6 +36,10 @@ on:
       - "requirements.txt"
       - "pyproject.toml"
 
+concurrency:
+  group: test-api-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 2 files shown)


goneall:
It looks like the API tests are hanging.

Here's the log file - it has been stuck on this last line for several minutes: 
[log.txt](https://github.com/user-attachments/files/31337504/log.txt)

The following error may be the root cause, but I'm not sure why it is hanging rather than failing:

```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "ClassLoader.java", line 525, in java.lang.ClassLoader.loadClass
java.lang.java.lang.ClassNotFoundException: java.lang.ClassNotFoundException: org.spdx.spreadsheetstore.SpreadsheetException
...
```

PR #720, opened 2026-08-22
harden: disable external XML entity processing in formatxml.py...

anupamme:
## Summary
Harden input handling in `src/app/formatxml.py` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410 |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410` |
| **File** | `src/app/formatxml.py:87` |
| **Assessment** | Defensive hardening |

**Description**: Found use of the native Python XML libraries, which is vulnerable to XML external entity (XXE)
attacks. The Python documentation recommends the 'defusedxml' library instead. Use 'defusedxml'.
See https://github.com/tiran/defusedxml for more information.

## Threat Model Context

This is a web service - vulnerabilities in request handlers are directly exploitable by remote attackers.

## Changes
- `src/app/formatxml.py`
- `requirements.txt`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+defusedxml>=0.7.1,<1.0.0
 django-oauth-toolkit>=3.2.0,<4.0.0
 drf-spectacular>=0.27.0,<1.0.0
 drf-social-oauth2>=3.4.1,<4.0.0
--- a/src/app/formatxml.py
+++ b/src/app/formatxml.py
@@ -84,7 +84,7 @@ VERSION = '1.0'
 import argparse
 import logging
 import re
-import xml.etree.ElementTree as et
+import defusedxml.ElementTree as et
 
 NL = '\n'
 XML_PROLOG = """<?xml version="1.0" encoding="UTF-8"?>"""


bact:
I have look a bit at the status of `defusedxml`.
The library was recommended by the Python documentation itself, [but no longer](https://github.com/python/cpython/pull/135294/changes#diff-d6ce2d53bcf20763d2f58b11018a9bfe0ea5e0ccfa3a027922dede23f024cf32L137) since January this year (2026).

See the discussion of the removal here:
https://discuss.python.org/t/status-of-defusedxml-and-recommendation-in-docs/34762/22

It is mainly because the library last updated was in 2021 and modern Python XML stdlib (since Python 3.11) already handle XML security issues of concerns. spdx-online-tools has recently up the floor for Python to 3.12, so I think it is not necessary to introduce to new dependency.

anupamme:
Good catch. I've reverted the defusedxml change (see 43386a4). Introducing it isn't warranted here:

  - defusedxml is no longer recommended. The Python docs dropped the recommendation in Jan 2026, and the library is effectively unmaintained (last release 2021). Discussion: https://discuss.python.org/t/status-of-defusedxml-and-recommendation-in-docs/34762/22
  - The stdlib is already safe on our Python floor (3.12). This project targets 3.12 everywhere (README, both Dockerfiles, all CI, pyproject.toml), and on 3.12:
    - `xml.etree.ElementTree` does not resolve external entities → no XXE / SSRF / local-file disclosure (it never did).
    - The "billion laughs" / entity-expansion DoS is mitigated by default by the modern expat (libexpat ≥ 2.4.0) bundled with Python 3.12, so et.parse() on untrusted input is protected without defusedxml.

  So the change was a no-op for security while adding an unmaintained dependency. I've reverted to `xml.etree.ElementTree` and added inline # nosec B405/B314 annotations with a justification comment so the SAST scanner doesn't re-flag and re-trigger an automated "fix."

bact (approved):
No changes in actual code.
Just "# nosec" comments added for security linters and comments for human reviewer to understand why they are suppressed.

PR #722, opened 2026-08-27
Switch to Valkey (from Redis)

### LINKED ISSUE #721: Try Valkey in Docker
Compare Valkey option to Redis.

While one of the three licenses of Redis is OSI-approved (AGPL), it is less compatible with Apache-2.0.

The more permissive license of Valkey (BSD 3-Clause) is considered more compatible Apache-2.0.

Valkey new hashtable design also reduces memory usage (but maybe negligible for our scale).

bact:
- Replace `redis` with `valkey` in requirements and codebase
- Update Docker Compose and CI images to `valkey:8-alpine`
- Add `SPDX_VALKEY_HOST` with `SPDX_REDIS_HOST` fallback
- Preserve RESP2 serialization protocol across connections
- Update docs

To fix #721


src/config/secret.py
```diff
@@ -63,6 +63,12 @@ def getOauthToolKitAppSecret():
 def getAuthCode():
     return os.environ.get(key="AUTH_CODE")
 
+def getValkeyHost():
+    host = os.environ.get(key="SPDX_VALKEY_HOST") or os.environ.get(key="SPDX_REDIS_HOST") or "localhost"
+    os.environ.setdefault("SPDX_REDIS_HOST", host)
+    return host
```
rtgdk, line 71: Mutating `os.environ` as a side effect of a getter is unexpected and can cause test pollution (env state leaking across test cases). Since all callers have been migrated to `getValkeyHost()`, this back-fill of `SPDX_REDIS_HOST` serves no purpose — I would remove this line.

requirements.txt
```diff
@@ -21,3 +20,5 @@ social-auth-app-django>=5.9.0,<6.0.0
 social-auth-core>=4.9.1,<5.0.0
 spdx-python-model>=0.0.4,<1.0.0
 spdx-tools>=0.8.5,<1.0.0
+valkey>=6.0.0,<7.0.0
+
```
rtgdk, line 24: Nit: `<7.0.0` is a reasonable conservative pin, but the `valkey` Python client and the Valkey server are versioned independently, so this may need a manual bump sooner than expected. A short comment noting the intent (e.g. "pin to known-compatible major; revisit when v7 ships") would help future maintainers.

README.md
```diff
@@ -284,23 +286,23 @@ can be found in the [wiki][rest-api].
 
 ### Redis serialization protocol version
 
-All `redis.StrictRedis(...)` connections in this codebase explicitly
+All `valkey.StrictRedis(...)` connections in this codebase explicitly
 pass `protocol=2` (RESP2) instead of relying on the client's default.
-The Redis server version used in our Docker image does not support RESP3,
-and the mismatch fails silently - the "check license" feature always returns
+The serialization protocol version ensures consistent behavior across servers,
```
rtgdk, line 292: The original text called out specifically that the Docker image's Redis version did not support RESP3, which was the concrete reason `protocol=2` is hardcoded. The new wording is vaguer. Since Valkey 8 *does* support RESP3, the reason has changed — worth adding a sentence explaining the current rationale (e.g. staying compatible with older Valkey/Redis instances that operators might still run).

rtgdk (approved):
Overall this is a clean, well-scoped migration — the backward-compatibility handling (env var fallback, function aliases) is thoughtful and the error-handling improvements in `isValkeyAvailable` are a nice bonus. A few small suggestions inline.

README.md
bact: Fixed.

src/config/secret.py
bact: `SPDX_REDIS_HOST` removed.

goneall (approved):

PR #725, opened 2026-09-20
Address the fork by its actual name when opening a pull request

### LINKED ISSUE #430: License XML editor assumes fork is named license-list-XML
When using the "License XML Editor" "Submit Changes" feature, if the user's fork of the license-list-XML repo is not named precisely `license-list-XML`, the submission will fail with a very generic error message.

The problem is in the [makePullRequest](https://github.com/spdx/spdx-online-tools/blob/43e330c70d4444e6cbf361c64b2fc0536aa6441f/src/app/utils.py#L92) code.

Eljees:
Fixes #430

### The defect

`makePullRequest` in `src/app/utils/__init__.py` addressed the contributor's fork as `repos/{username}/{settings.LICENSE_REPO_NAME}` (or `NAMESPACE_REPO_NAME` for namespaces) in every call it makes: updating the fork from upstream, reading the ref of `main`, listing branches, creating the branch and committing the files.

Those settings name the **upstream** repository. A fork does not have to carry that name. GitHub itself appends `-1`, `-2` and so on when the account already holds a repository with the default name, and the user can rename it afterwards. Once that happens, every one of those calls points at a repository that does not exist and the submission ends with the generic *"Error occured while ... Please try again later or contact the SPDX Team"*, which gives neither the user nor the maintainers a clue about the real cause.

### The change

Take the fork's real name from the data the function already fetches: the `name` of the entry in `GET /forks` whose owner is the user, and the `name` from the creation response when the fork is made on the spot. Everything downstream uses that value.

### What I ran

Python 3.12 with `requirements.txt` + `requirements-dev.txt`, valkey 8 as in the App tests workflow, `python manage.py test tests.app.test_make_pull_request -v 2`:

- control, the file as it is on `main`: **1 test, 1 failure** — `AssertionError: Lists differ: ['https://api.github.com/repos/forktestuser/...'] != []`, that is, the call went to the settings-derived name while the fork is called something else
- with the change: **1 test, OK**

The new test drives `makePullRequest` with `requests.get`/`requests.post` doubled so that only the fork's real name answers and any other name returns 404, then asserts that no request was addressed to `settings.LICENSE_REPO_NAME` under the user's account.


bact (approved):
Looks reasonable.

goneall (approved):

